PR #40, opened 2016-11-30
Remove non-existant dependency on configure-step

wxmerkt:
FYI @VladimirIvan: the dependency does not exist as this file is generated during the configure-step. Removing it reduces warnings during build time.


VladimirIvan:
ok

Issue #41, opened 2016-12-05
Support for different robot_description ROS params


wxmerkt:
- Historically, a ``<Urdf>`` node underneath ``<Kinematica>`` specified the rosparam on which the URDF was. These are still in all initialisation files, but unused.
- Instead, ``robot_description`` is hard-coded in the ``Scene.cpp`` initialisers.
- In some cases, e.g. for the husky, the robot_description channel contains another description than the one we'd like. [We're seeing rviz-crashes due to the differences for instance]
- We'd like to be able to set the param on which to look for the description as one of the initialisation flags, as it used to be. E.g. ``/exotica/robot_description``

Hence, 

a) would we like to reactivate use of the ``<Urdf>`` tag - or perhaps another tag or directly in a server parameter (where we now have packages and urdf file names inside packages)
b) would we like to delete all instances of the ``<Urdf>`` tag remnants and switch to something else?

cc @VladimirIvan

wxmerkt:
Proposal:

1. We'll add an optimal parameter ``RobotDescription`` to the Server.
2. If ``RobotDescription`` is found, it uses the param name defined there. If it's not, it falls back to ``robot_description``.

Does that sound fair?

PR #42, opened 2016-12-05
Adds support for different robot_description parameter name.

wxmerkt:
The parameter to use instead of the default robot_description ROS param can be set as a string EXOTica-Server parameter named ``<RobotDescription>``.

Addresses and resolves issue #41.

@VladimirIvan is this okay?


VladimirIvan:
Oh, the original version wasn't working the way I thought it did.
I will redo this in the future but I'm happy to stick with this version for now.
The feature is definitely needed.

Issue #48, opened 2017-05-04
Use movegroups from SRDF


VladimirIvan:
The joint group used for planning is currently defined in the initializer of the Scene.
The same information is already specified in the SRDF though.
Using the SRDF set of joints would make the initializer shorter and put the joint group in one unique place (the SRDF). This will also make it easier to use the MoveIt features more directly, e.g. setting the joint state, joint state indexing, joint limits, setting the unspecified joints (robot state), ...

VladimirIvan:
Fixed in #53

PR #53, opened 2017-05-08
Use joint groups from SRDF

VladimirIvan:
Solves #48
 * The scene now requires the joint group to use. This group has to be defined in the SRDF.
 * Kinematica uses this group name to initialize itself. No more information is required. The kinematica initializer was therefore removed.
 * Base joint, root frame, active joint names and base type are parsed directly from the model. Base is always movable (assuming the root joint is not fixed) and non-specified joints are set to zero by default (to be addressed in the future).

Additionally, ```bool``` return values in Kinematica have been replaced with exceptions. Doxygen style comments inside functions have been replaced with regular c++ comments.


VladimirIvan:
Merge #54 first

PR #54, opened 2017-05-09
Depricated code deletion

VladimirIvan:
* Removes initialization directly from XML (see new initialization method)
 * Removes JSON re-initialization (to be replaced with new init. and/or python wrappers)
 * Removed deprecated tools 
 * Removed dependencies on tinyxml and rapidjson
 * Removes exotica_distro and eclipse formatting definitions


exotica/src/Scene.cpp
```diff
@@ -72,113 +72,6 @@ namespace exotica
     //TODO
   }
 
-  void CollisionScene::reinitialise()
```
wxmerkt: This is where collision checking functionality got broken - cf. #98

PR #58, opened 2017-06-01
Initializer fix and Compatibility with Ubuntu Xenial 16.04 LTS and Kinetic Kame

### LINKED ISSUE #57: Initializers don't work on Kinetic
The CMake macro logic for the initializers doesn't work with catkin tools if configured to _not_ use an install directory as the files are generated in the devel and not the source tree whereas ``${catkin_INCLUDE_DIRS}`` will return the source tree path. Generating into the source tree is not a good CMake option either. 

Questions thus are how to handle non-install configured workspaces?

[The issue arises by just building exotica - when building the examples it will fail as it cannot find the generated {something}Initializer.h file]

wxmerkt:
This PR brings in a range of updates:

1. It fixes the initializers for non-install spaces and thereby properly resolves #57 
2. It removes some non-required dependencies that were spewing warnings
3. It makes Exotica compatible with 16.04 and Kinetic Kame.

Fully tested with both ``catkin_make_isolated`` and ``catkin build`` on Xenial + Kinetic. Travis confirmed that all changes work with Trusty + Indigo.

### DIFF
--- a/examples/exotica_examples/CMakeLists.txt
+++ b/examples/exotica_examples/CMakeLists.txt
@@ -9,7 +9,6 @@ find_package(catkin REQUIRED COMPONENTS
 )
 
 catkin_package(
-
 )
 
 include_directories(
@@ -18,7 +17,6 @@ include_directories(
 )
 
 add_executable(ExoticaCore src/core.cpp)
-
 target_link_libraries(ExoticaCore
   ${catkin_LIBRARIES}
 )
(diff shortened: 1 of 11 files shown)


VladimirIvan:
Can you move the:
```
install(DIRECTORY include/ DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
install(DIRECTORY ${CATKIN_DEVEL_PREFIX}/include/ DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
```
into the ```AddInitializer.cmake``` ?

I'm supportive of removing boost all around in favour of std. Especially since ROS is going the same direction.

Issue #49, opened 2017-05-04
Remove task definitions entirely


VladimirIvan:
After implementing #46, the problem can hold everything necessary for handling a task map within its scope. The task definition is then redundant. The precision parameter (rho) can moved directly into the problem together with any time parametrization and other specifics.

Action: remove task definition class and move some of its functionality into the problems (where appropriate).

VladimirIvan:
This would be easiest when done after #44

VladimirIvan:
Fixed in #59

Issue #47, opened 2017-05-04
FK to use KDL as output


VladimirIvan:
The kinematic solver currently outputs Eigen vectors and matrices (for eff. poses and Jacobians respectively). The solver does use KDL internally and the task maps can easily use the KDL as input (```KDL::Frame``` and ```KDL::Jacobian```). This would make the kinematic solver simpler and the operations in task map more transparent (e.g. dealing with frames instead of indexing into matrices).
The task map will have to convert the output into a Eigen::Vector/Matrix in the end but this will make the operation explicit (unlike doing it implicitly in the kinematic solver).

VladimirIvan:
This will require extra work on Kinematica. It currently only supports positions and position Jacobians. Those in batch using only minimal updates. The positions are just extracted from KDL::Frame but the Jacobian is not a full geometric KDL::Jacobian. The solver has to updated, ideally based on KDL::TreeJntToJacSolver.

VladimirIvan:
Fixed in #59

Issue #44, opened 2017-05-04
Backtrack definition/map external data allocation


VladimirIvan:
The inputs/outputs/parameters for task maps and task definitions are currently allocated by the solver (see [here](https://github.com/openhumanoids/exotica/blob/master/exotations/solvers/ik_solver/src/ik_solver/IKSolver.cpp#L163)).
This makes the code hard to read and easy to introduce difficult to track memory access issues.
Each class should hold its own data or get it passed as an argument in a function.

To remove this feature (which was mistakenly added in the past):
 * Make task map output the task space value on update.
 * Move map/definition parameters back to the map/definition.
 * Remove ```registerPhi```, ```registerJacobian```, ...
 * Update existing solvers/problems.

VladimirIvan:
This would be easiest when done after #47

VladimirIvan:
Fixed in #59

PR #59, opened 2017-06-16
Kinematics and Planning problem rework

### LINKED ISSUE #46: Clean up IK solver
The IK solver has been accumulating a lot of code.
The implementation should only be a couple of lines long.

Unused code should be removed. Useful code should be simplified. Variables and methods have to named more appropriately.

VladimirIvan:
This PR changes the following core feature of EXOTica:
 - Phi, Jacobian, Goal, Rho, ... is now stored in the specific planning problem.
 - Phi and Jacobian are returned from task map update method via arguments.
 - ```SweepFlux``` task map was removed (it's not currently compatible with the previous change).
 - Task definitions were removed, thir functionality got merged into the planning problem. ```get/setRho```, ```get/setGoal``` now exist in the problem classes.
 - Planning problem can have zero task maps defined.
 - The scene now uses joints from the SRDF move group (including the floating base). Controlled joints are now specified in the SRDF not in the XML!
 - Kinematic solver now computes ```KDL::Frame``` and ```KDL::Jacobian``` arrays (task maps use these directly).
 - Kinematic requests to the solver now include full relative transforms: from base to link.
 - Kinematic solver now publishes frames to TF when in debug mode (call ```publishFrames``` when not in debug mode)
 - Limb initializer has been replaced with a Frame initializer (to accommodate the above change).
 - Task map test has been added.
 - IK solver has been simplified.
 - IK solver now takes an optional ```qNominal``` pose parameter via the ```UnconstrainedEndPoseProblem```.
 - XML files with comments now load correctly.
 - Unused code was removed

This PR fixes #46, #47, #49 and #44.


examples/exotica_examples/src/manual.cpp
```diff
@@ -32,61 +32,43 @@ void run()
 
     // Assign the problem to the solver
     any_solver->specifyProblem(any_problem);
+    UnconstrainedEndPoseProblem_ptr my_problem = boost::static_pointer_cast<UnconstrainedEndPoseProblem>(any_problem);
+
 
     // Create the initial configuration
-    Eigen::VectorXd q = Eigen::VectorXd::Zero(any_problem->scene_->getNumJoints());
+    Eigen::VectorXd q = Eigen::VectorXd::Zero(any_problem->getScene()->getNumJoints());
```
wxmerkt: Here you use ``getScene()->getNumJoints()`` while below you use ``->N``

examples/exotica_examples/src/xml.cpp
```diff
@@ -23,61 +23,43 @@ void run()
 
     // Assign the problem to the solver
     any_solver->specifyProblem(any_problem);
+    UnconstrainedEndPoseProblem_ptr my_problem = boost::static_pointer_cast<UnconstrainedEndPoseProblem>(any_problem);
 
     // Create the initial configuration
-    Eigen::VectorXd q = Eigen::VectorXd::Zero(any_problem->scene_->getNumJoints());
+    Eigen::VectorXd q = Eigen::VectorXd::Zero(my_problem->N);
```
wxmerkt: Here it is ``->N`` - is this because there may be uncontrolled joints?

exotations/solvers/aico/src/AICOsolver.cpp
```diff
@@ -82,7 +75,7 @@ namespace exotica
     myfile << "Control";
     for (auto s : taskNames)
       myfile << " " << s;
-    for (int t = 0; t < T + 1; t++)
+    for (int t = 0; t < T; t++)
```
wxmerkt: LGTM - Was this an off-by-one bug and how did it manifest?

exotations/solvers/aico/src/AICOsolver.cpp
```diff
@@ -265,12 +163,9 @@ namespace exotica
...
   void AICOsolver::initMessages()
   {
-    if (prob_ == nullptr)
-    {
-      throw_named("Problem definition is a NULL pointer!");
-    }
+    if (prob_ == nullptr) throw_named("Problem definition is a NULL pointer!");
     // TODO: Issue #4
-    n = prob_->getScene()->getNumJoints();
+    n = prob_->N;
```
wxmerkt: Same as above - which one do we prefer?

exotica/CMakeLists.txt
```diff
@@ -51,20 +51,8 @@ include(GetGitRevisionDescription)
...
-  PlanningProblem
-  Scene
-  TaskDefinition
-  UnconstrainedTimeIndexedProblem
-  UnconstrainedEndPoseProblem
-  SamplingProblem
-  TaskSqrError
-  TaskTerminationCriterion
-)
+AddInitializer(TaskMap Object Frame MotionSolver PlanningProblem Scene
```
wxmerkt: Why this formatting change? It's easier to add and remove with the open array without changing the whole list of initializers

exotica/include/exotica/Factory.h
```diff
@@ -7,9 +7,9 @@
  * Redistribution and use in source and binary forms, with or without 
  * modification, are permitted provided that the following conditions are met: 
  * 
- *  * Redistributions of source code must retain the above copyright notice, 
+ *  * Redistributions of source code must retain the aBaseClassve copyright notice,
```
wxmerkt: this should be reverted

exotica/include/exotica/Factory.h
```diff
@@ -7,9 +7,9 @@
...
  *    this list of conditions and the following disclaimer. 
- *  * Redistributions in binary form must reproduce the above copyright 
+ *  * Redistributions in binary form must reproduce the aBaseClassve copyright
```
wxmerkt: same - case-sensitive find and replace would resolve it

exotica/include/exotica/Scene.h
```diff
@@ -230,227 +225,31 @@ namespace exotica
...
-       */
-      void getPoses(const std::vector<std::string> & names,
-          std::vector<KDL::Frame> & poses);
-
-      /**
-       * @brief getRobotModel Returns the robot model
-       * @return Robot model
-       */
       robot_model::RobotModelPtr getRobotModel();
-
```
wxmerkt: Why did you remove all doxygen? We may have to re-add that later on

exotations/solvers/aico/src/AICOsolver.cpp
VladimirIvan: It would have caused a memory issue silently. It didn't cause any problems in practice as far as I can remember so I guess the code was not executed very often (or at all) since the bug was introduced.

exotica/include/exotica/Factory.h
VladimirIvan: Yep. Fixed now

exotica/include/exotica/Scene.h
VladimirIvan: I have removed Doxygen where the method/variable name was explicit enough.
The code was becoming hard to edit with all the comments.

PR #69, opened 2017-07-25
Goal setting for OMPL fixed

cam586:
Bug for OMPL goal setting fixed


VladimirIvan:
Merged in #70 directly into master

PR #72, opened 2017-08-03
Python wrapper

VladimirIvan:
Wraps core exotica and task_maps in python.

Python related:
- Added ```pybind11``` package
- Added ```exotica_py``` package (wrapping the core classes)
- Added python wrapping to ```task_map``` package
- Added python examples mirroring ``exotica_examples```
- Added python3 compatibility (links with python3 libraries but does not work with ros python packages, hence only for standalone use within python)

Other:
- Added map to vector conversion
- Moved type name demangling to tools 
- Fixed ```Property::set```
- Added ```Server::destroy``` method
- moved task_map header files from ```include``` to ```include/task_map```
- Changed ```task_map``` package format to ```2```


exotations/task_maps/task_map/package.xml
```diff
@@ -13,10 +13,9 @@
   <url type="website">https://bitbucket.org/IPAB-SLMC/exotica</url>
```
wxmerkt: Maybe for other PR: Need to update the URLs

exotica_py/package.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>exotica_py</name>
```
wxmerkt: Shall we do the rename to pyexotica now or what speaks against it?

exotica_py/src/Exotica.cpp
```diff
@@ -0,0 +1,508 @@
...
+
+
+PYBIND11_MODULE(exotica_py, module)
+{
+    //Setup::Instance();
+
+    module.doc() = "Exotica Python binding";
+
+    int argc = 0;
+    ros::init(argc, nullptr, "ExoticaPyDummyNode");
```
wxmerkt: We should update both ``module.doc()`` (L358) and the ROS init before merging

exotica_py/src/Exotica.cpp
```diff
@@ -0,0 +1,508 @@
...
+    motionSolver.def("solve", [](std::shared_ptr<MotionSolver> sol, Eigen::VectorXd q0){return Solve(sol, q0);}, "Solve the problem", py::arg("q0"));
+    motionSolver.def("getProblem", &MotionSolver::getProblem, py::return_value_policy::reference_internal);
+
+    py::class_<PlanningProblem, std::shared_ptr<PlanningProblem>, Object> planningProblem(module, "PlanningProblem");
+    planningProblem.def("getTasks", &PlanningProblem::getTasks, py::return_value_policy::reference_internal);
+    planningProblem.def("getScene", &PlanningProblem::getScene, py::return_value_policy::reference_internal);
+    planningProblem.def("__repr__", &PlanningProblem::print, "String representation of the object", py::arg("prepend")=std::string(""));
+
+    // Problem types
+    py::module prob = module.def_submodule("Problems","Problem types");
```
wxmerkt: As I see it, every module needs to be individually added to this wrapper - i.e. the support at the moment is limited to the ``UnconstrainedTimeIndexedProblem`` and ``SamplingProblem`` (main problem classes nonetheless) - is that correct?

exotica_py/src/Exotica.cpp
```diff
@@ -0,0 +1,508 @@
...
+    kinFrameReq.def(py::init<std::string, KDL::Frame, std::string, KDL::Frame>());
+    kinFrameReq.def(py::init<std::string, KDL::Frame, std::string>());
+    kinFrameReq.def(py::init<std::string, KDL::Frame>());
+    kinFrameReq.def(py::init<std::string>());
+    kinFrameReq.def_readwrite("frameALinkName", &KinematicFrameRequest::FrameALinkName);
+    kinFrameReq.def_readwrite("frameAOffset", &KinematicFrameRequest::FrameAOffset);
+    kinFrameReq.def_readwrite("frameBLinkName", &KinematicFrameRequest::FrameBLinkName);
+    kinFrameReq.def_readwrite("frameBOffset", &KinematicFrameRequest::FrameBOffset);
+
+    py::class_<KDL::Frame> kdlFrame (module, "KDLFrame");
```
wxmerkt: We should add example code for the frame requests - this is useful stuff!

pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+include(pybind)
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+
+file(MAKE_DIRECTORY include)
+catkin_package(
+  INCLUDE_DIRS include pybind11/include ${EIGEN3_INCLUDE_DIR}
```
wxmerkt: Is there a find_package(Eigen3) required given that it's a standalone package?

exotations/task_maps/task_map/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ add_definitions(-DEXOTICA_DEBUG_MODE)
```
VladimirIvan: It's a ros standard directory structure that way. It should have been done when the package was created in the first place.

exotations/task_maps/task_map/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ add_definitions(-DEXOTICA_DEBUG_MODE)
 
 find_package(catkin REQUIRED COMPONENTS
   exotica
+  exotica_py
```
VladimirIvan: You don't need ```task_map``` to build ```exotica_py```. The package is a plugin so you can still use it without linking to it. It scales if you add new packages. Otherwise you would have to make ```exotica_py``` depend on all packages in c++. This way you only import them in python when needed.
```task_map_py``` only adds specification of new classes in python but you can still use the abstract interface of ```TaskMap``` (defined in ```exotica```). In most cases that is all you want to do anyway.

File: exotations/task_maps/task_map/CMakeLists.txt [hunk 9]
```diff
@@ -49,6 +50,9 @@ target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
 add_dependencies(${PROJECT_NAME} ${PROJECT_NAME}_initializers)
 
+pybind_add_module(${PROJECT_NAME}_py MODULE src/task_map_py.cpp)
```
VladimirIvan: same as above

exotica_py/package.xml
VladimirIvan: http://wiki.ros.org/ROS/Patterns/Conventions

exotica_py/src/Exotica.cpp
VladimirIvan: To what?
```module.doc()``` is describing the module and node name is referring to just a dummy node required to run ROS commands internally. Open to suggestions

File: exotica_py/src/Exotica.cpp (hunk 4 above)
Comment (VladimirIvan) [inline exotica_py/src/Exotica.cpp]: Only code in the ```exotica``` package has to be added and only when we want it exposed.
Sub-modules are only for creating nicer looking hierarchies within python.
Other packages can create their own modules or they can add classes tho the main module. This is supported dynamically [across different libraries](http://pybind11.readthedocs.io/en/stable/advanced/misc.html#partitioning-code-over-multiple-extension-modules).

File: exotica_py/src/Exotica.cpp (hunk 5 above)
Comment (VladimirIvan) [inline exotica_py/src/Exotica.cpp]: Agreed. We also need tests.

File: pybind11/CMakeLists.txt (hunk 6 above)
Comment (VladimirIvan) [inline pybind11/CMakeLists.txt]: It gets called inside one of the cmake files included from pybind11 which also gets called in dependant packages via ```pybind.cmake```, so I'm guessing not.

File: exotations/task_maps/task_map/package.xml [hunk 10]
```diff
@@ -13,10 +13,9 @@
...
 
   <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>exotica</build_depend>
-  <build_depend>geometry_msgs</build_depend>
-  <run_depend>exotica</run_depend>
-  <run_depend>geometry_msgs</run_depend>
+  <depend>exotica</depend>
+  <depend>exotica_py</depend>
```
VladimirIvan: see the comment on the CMakeLists

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "pybind11/pybind11"]
```
VladimirIvan: In the current format yes. But I want to put the package description directly into the original repo. It will have to be a submodule then.

exotations/task_maps/task_map/package.xml
VladimirIvan: done

.gitmodules
wxmerkt: Okay, in that case I would propose waiting to see whether they accept the catkin PR. Otherwise let's make it an add external project and move on

exotations/task_maps/task_map/CMakeLists.txt
wxmerkt: I just checked the content and forgot that dmesh_ros was also in that package, makes absolute sense.

exotations/task_maps/task_map/CMakeLists.txt
wxmerkt: My comment was targeting the other way around - now ``task_map`` depends on ``exotica_python`` being built. Makes sense, but we should make the building of the Python bindings conditional, i.e. only call ``pybind_add_module`` if we know Python is present and has fully been built?

exotations/task_maps/task_map/CMakeLists.txt
wxmerkt: Resolved, discussion now above

exotations/task_maps/task_map/package.xml
wxmerkt: Resolved, discussion now above

exotica_py/package.xml
wxmerkt: These naming conventions relate to package names, not Python module names (PEP-8 supports underscores if they both are whole words and readability is improved). ``py`` is not a whole word. After thinking things, through, perhaps the following is the best:

- ROS package ``exotica_python`` and Python module name ``exotica`` (drop the py or python from the module name - as a suffix it looks very ugly, as a prefix would look better)

Perhaps let's have a chat F2F over lunch?

exotica_py/src/Exotica.cpp
wxmerkt: Semantically, "bindings". Perhaps exotica_python_node? Dummy just doesn't sound nice and descriptive.

exotica_py/src/Exotica.cpp
wxmerkt: Thanks for the pointer - fantastic.

wxmerkt:
Thanks for the update, just added a few more comments. Perhaps let's sort this out over lunch and get it merged?

exotations/task_maps/task_map/CMakeLists.txt
VladimirIvan: That would be the way for pure CMake but it's rarely used in Catkin. We could do what you suggest by pulling the ```task_map_py``` into a separate package. It would be bit more tricky because ```libtask_map.so``` is a plugin library and would have to loaded as such, since you cannot link to a library and use it as a plugin in the same binary. This would then allow you remove/ignore both ```exotica_py``` and any extension wrappers.

wxmerkt: Let's leave it as is.

exotica_py/src/Exotica.cpp
VladimirIvan: updated to ```Exotica Python wrapper``` and ```exotica_python_node```

wxmerkt:
Ubuntu 16.04 + Kinetic issue when running the examples: 

```
[EXOTica]: Skipping unsupported type 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >'
```

exotations/task_maps/task_map/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ add_definitions(-DEXOTICA_DEBUG_MODE)
```
VladimirIvan: Can you revert the bash line in ```example_aico.py``` back to ```python``` instead of ```python3```?

PR #78, opened 2017-08-07
Fixed install target

VladimirIvan:
- Added missing pluxing xml file install commands (task maps and solvers)
- Added missing launch file commands (examples and exotica_python)
- Added python example install commands
- Fixed pybind11 install commands


exotations/task_maps/task_map/CMakeLists.txt
```diff
@@ -53,13 +53,14 @@ target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 add_dependencies(${PROJECT_NAME} ${PROJECT_NAME}_initializers)
 
 pybind_add_module(${PROJECT_NAME}_py MODULE src/task_map_py.cpp ${SOURCES})
-install(TARGETS ${PROJECT_NAME}_py LIBRARY DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION})
+install(TARGETS ${PROJECT_NAME}_py LIBRARY DESTINATION ${CATKIN_GLOBAL_PYTHON_DESTINATION})
```
wxmerkt: Do we want to install to the global destination?

exotica_python/CMakeLists.txt
```diff
@@ -16,4 +16,6 @@ include_directories(
 )
 
 pybind_add_module(pyexotica MODULE src/Exotica.cpp)
-install(TARGETS pyexotica LIBRARY DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION})
+install(TARGETS pyexotica LIBRARY DESTINATION ${CATKIN_GLOBAL_PYTHON_DESTINATION})
+catkin_install_python(PROGRAMS scripts/example_aico.py scripts/example_ompl.py scripts/example_ik.py scripts/publish_trajectory.py DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
```
wxmerkt: While I agree with using ``catkin_install_python``, publish trajectory is a helper of the pyexotica module and not a standalone program so shouldn't be here.

exotations/task_maps/task_map/CMakeLists.txt
VladimirIvan: We have to. The ```*.so``` files have to in python path for them to be recognized as modules. The package destination is a subfolder of the global destination. All the python files go there and it requires a ```_init__.py``` to be recognized as a module.

exotica_python/CMakeLists.txt
VladimirIvan: Is there a simple way to include it if we put it somewhere else?
We may have to set up that subdirectory in the global path and add both the ```*.so``` and ```publish_trajectory.py``` into ```_init__.py```.

wxmerkt: Perhaps a submodule, i.e. ``from pyexotica.tools import publish_trajectory``. So ``pyexotica`` is the ``.so`` and ``tools`` a folder inside the module folder?

wxmerkt: Can we resolve this here before #79? Makes the other easier to merge :-)

VladimirIvan:
- Renamed the `*.so` to `pyexotica_impl```
- Created `setup.py` and `__init__.py` to set up a `pyexotica` module (same name as before)
- Moved `publishTrajectory.py` into the module

wxmerkt:
Nice, that should work. Common name for (back then C-based) modules that are wrapped with higher-level Python code as here would be with a leading underscore, i.e. ``_pyexotica`` - mind renaming it to that? {I got that from [here](http://dan.iel.fm/posts/python-c-extensions/) when I wrapped my other C-code}

Otherwise, awesome! Does that still work with the examples?

VladimirIvan:
Yes, it works with examples and in the install space too.

PR #79, opened 2017-08-08
pybind11 as external project

### LINKED ISSUE #73: pybind11 catkin wrapper
The catkin wrapper should either be renamed to ``pybind11_catkin`` to avoid clashes and use external project to checkout the source, or upstream the catkin related changes.

VladimirIvan:
- Renames ```pybind11``` to ```pybind11_catkin```
- Removes submodule ```pybind11/pybind11```
- ```pybind11``` is now and external setup from within ```pybind11_catkin```.
- Fixes #73 

Do not ```find_package``` the ```pybind11``` directly! Use ```pybind11_catkin``` instead. The external project has compatibility issue with cmake>3.0 due to interface library definitions. The catkin package gets around this and adds a convenience macro that should be used instead: ```pybind_add_module```.

To use different python version for wrapping set ```PYBIND_PYTHON_EXECUTABLE``` cmake variable to path to the python executable you want to use: ```-DPYBIND_PYTHON_EXECUTABLE=/usr/bin/python3```.


pybind11_catkin/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+find_package(catkin REQUIRED)
+
+include(ExternalProject)
+
+set(PYBIND_SRC ${CMAKE_BINARY_DIR}/pybind11_src)
+
+ExternalProject_Add(
+  pybind11
+  GIT_REPOSITORY "git@github.com:pybind/pybind11.git"
```
wxmerkt: This should be the HTTPS version - ``https://github.com/pybind/pybind11.git``. Otherwise will fail if you don't have a private key registered with GitHub on that machine (e.g. build server)

wxmerkt:
LGTM, will test again once #78 is merged

wxmerkt:
NB: Requires ``catkin clean`` to get the symlink to work, then works like a bliss.

PR #87, opened 2017-08-09
Bring Continuous Integration back online

wxmerkt:
Brings back continuous integration:

1. Now uses the packaged Eigen debian we recommend (we now require >3.2.7 for the Python bindings)
2. Fixes a PYTHONPATH bug that Travis recently introduced
3. Fixes dependency on ``moveit_ros_planning`` instead of ``moveit_ros_planning_interface`` [the latter had installation issues due to depending on the warehouse database]
4. Fixes a CMake error we hadn't caught without CI.

Still needs adding the Python tests and proper failure if tests fail. The Python tests won't work on Indigo as they all segfault due to an upstream roscomm issue.

### DIFF
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -1,10 +1,13 @@
 #!/bin/bash
 set -xe
-wget https://bitbucket.org/eigen/eigen/get/3.2.6.zip
-unzip 3.2.6.zip
-cd eigen-eigen-c58038c56923
-mkdir build
-cd build
-cmake ..
-sudo make install
-cd ../..
+wget http://terminator.robots.inf.ed.ac.uk/apt/libeigen3-dev.deb
+sudo dpkg -i libeigen3-dev.deb
+rm libeigen3-dev.deb
+# wget https://bitbucket.org/eigen/eigen/get/3.2.6.zip
+# unzip 3.2.6.zip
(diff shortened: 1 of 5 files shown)


VladimirIvan:
I'm using gcc5.
Can you confirm if this is this a c++11 issue or just gcc issue in general?
gcc5 defaults to c++11 so I might have missed a flag somewhere.

wxmerkt:
The failure appears to be due to a gcc-4.8.x bug in combination with using optimisation (on Travis ``-O2`` by default). Easiest solution for now should be to upgrade the gcc compiler suite on Travis to >=4.9. This is what I am testing right now.

- [gcc bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58528)
- In the future we can test against multiple gcc versions (via a [matrix](https://docs.travis-ci.com/user/languages/c/))

wxmerkt:
It was indeed the gcc version and upgrading to 4.9 resolved it, cf. [here](https://travis-ci.org/openhumanoids/exotica/builds/262884671). I am now making sure the builds show as failed when the tests fail and will then merge and we've got CI back :)

wxmerkt:
All tests passed and CI is back up: https://travis-ci.org/openhumanoids/exotica/builds/262900412

PR #88, opened 2017-08-10
added indexing error handling

### LINKED ISSUE #76: Error handling and helpful messages
Error handling, e.g. catch if a link does not exist and give helpful warnings/messages. (Longer standing issue to track issues as they pop up)

- [ ] Kinematic frame requests - report if a frame does not exist rather than having the ``map.at(`` throw a cryptic error - [here](https://github.com/openhumanoids/exotica/blob/e4014609df8951174c9e2f7097d6a3d1a80fafa2/exotica/src/KinematicTree.cpp#L233-L241)

cam586:
Error handling added 

Resolves #76

### DIFF
--- a/exotica/src/KinematicTree.cpp
+++ b/exotica/src/KinematicTree.cpp
@@ -236,12 +236,25 @@ std::shared_ptr<KinematicResponse> KinematicTree::RequestFrames(const Kinematics
         if(request.Frames[i].FrameALinkName=="")
             Solution->Frame[i].FrameA = Root;
         else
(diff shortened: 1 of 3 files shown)


wxmerkt:
Nice, would be interesting to see whether there is a time/performance penalty due to the try-catch block or whether a find/checking index-statement would be faster.

VladimirIvan:
All those functions only get called on initialization. Solve time won't be affected at all. Even in the initialization, the `at` method will probably take more time than what try/catch introduces.

PR #90, opened 2017-08-18
Minor changes

VladimirIvan:
- Removes ros node dependent code from AICO.
- Adds `Timer` class wrapping chrono.
- Removed internal exotica methods/classes from pyexotica (they are not meant be used through the wrapper in the first place).
- Added `setModel` convenience function for setting URDF and SRDF file path from python.
- Added path parsing tools:
    - Paths relative to package can be now specified using `{package_name}/resources/...`.
    - Scene parses paths when loading URDF, SRDF.
    - URDF/SRDF files are now specified in the XML config. Demos use these. `robot_description` is used when UDRF/SRDF files are not specified.
    - XML loader parses package paths.
- Python wrappers for common tools added.
- Moved resources from `exotica_examples` to `exotica` package.


exotica/include/exotica/Tools.h
```diff
@@ -81,6 +82,12 @@ namespace exotica
   }
 
   std::string getTypeName(const std::type_info& type);
+
+  std::string parsePath(const std::string& path);
+
+  std::string loadFile(const std::string& path);
+
+  bool exists(const std::string& path);
```
wxmerkt: ``exists`` is not necessarily clear - perhaps ``fileExists`` or ``folderExists`` or does it check either?

exotica/src/Tools.cpp
```diff
@@ -37,6 +37,8 @@
 #include <iostream>
 #include <typeinfo> //!< The RTTI Functionality of C++
 #include <cxxabi.h> //!< The demangler for gcc... this makes this system dependent!
+#include <regex>
```
wxmerkt: This will definitely make our gcc requirement >4.9, which we already require anyways to avoid segfaults from the wrapper. Want to flag it up either way (non-14.04 standard compiler).

exotica_python/src/Exotica.cpp
```diff
@@ -476,16 +496,6 @@ PYBIND11_MODULE(_pyexotica, module)
     py::class_<KinematicTree, std::shared_ptr<KinematicTree>> kinematicTree(kin, "KinematicTree");
     kinematicTree.def("publishFrames", &KinematicTree::publishFrames);
 
-    py::class_<KinematicFrameRequest, std::shared_ptr<KinematicFrameRequest>> kinFrameReq(kin, "KinematicFrameRequest");
```
wxmerkt: Wondering whether these should be removed? May be useful for debugging or when used in combination with other tools

VladimirIvan: These are classes only used internally. I would like to keep the interface clear. 
It has very limited use for debugging. If there is a problem with a frame request, it should have been dealt with in c++. You don't even get to construct those classes when you are defining new task maps.

exotica/src/Tools.cpp
VladimirIvan: Good observation.
Let's keep moving towards 16.04 but make a list of dependencies for now.

VladimirIvan:
I can't recreate the test failure on my PC any tips?

I moved the resources to avoid cross dependencies between exotica_python and exotica_examples. Also, I wanted start creating resources for test. Both issues can resolved in other was. May revisit this later.

wxmerkt:
I can't access the logs from this hotel (blocked), worked in the other. Will have a look later and update this thread.

exotica/src/Tools.cpp
```diff
@@ -146,4 +148,47 @@ namespace exotica
...
+  {
+      std::string ret = path;
+      std::smatch matches;
+      std::regex_search(ret, matches, std::regex("\\{([^\\}]+)\\}"));
+      for(auto& match : matches)
+      {
+          std::string package = match.str();
+          if(package[0]=='{') continue;
+          std::string package_path = ros::package::getPath(package);
+          if(package_path=="") throw_pretty("Unknown package '"<<package<<"'");
```
wxmerkt: This is the line that throws on CI. CI uses gcc-4.9 but perhaps there are still issues with the regex implementation in that version.

wxmerkt:
The issue is that in ``loadModel`` you are trying to do a ``pathExists`` on a XML-string:

gdb traceback:
```
[EXOTica]: Parsing EXOTica paths...
terminate called after throwing an instance of 'exotica::Exception'
  what():  

Program received signal SIGABRT, Aborted.
0x00007fffeb96ec37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  0x00007fffeb96ec37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007fffeb972028 in __GI_abort () at abort.c:89
#2  0x00007fffec28a415 in __gnu_cxx::__verbose_terminate_handler() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007fffec288206 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007fffec288251 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007fffec288493 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7b3e974 in exotica::parsePath (
    path="<robot name=\"test_robot\"><link name=\"base\"><visual><geometry><cylinder length=\"0.3\" radius=\"0.2\"/></geometry><origin xyz=\"0 0 0.15\"/></visual></link><link name=\"link1\"> <visual><geometry><cylinder len"...)
    at /home/wxm/tmp_ws/src/exotica/exotica/src/Tools.cpp:160
#7  0x00007ffff7b3efa6 in exotica::pathExists (
    path="<robot name=\"test_robot\"><link name=\"base\"><visual><geometry><cylinder length=\"0.3\" radius=\"0.2\"/></geometry><origin xyz=\"0 0 0.15\"/></visual></link><link name=\"link1\"> <visual><geometry><cylinder len"...)
    at /home/wxm/tmp_ws/src/exotica/exotica/src/Tools.cpp:192
#8  0x00007ffff7b73723 in exotica::Server::loadModel (this=this@entry=0x682030, name="robot_description", 
    urdf="<robot name=\"test_robot\"><link name=\"base\"><visual><geometry><cylinder length=\"0.3\" radius=\"0.2\"/></geometry><origin xyz=\"0 0 0.15\"/></visual></link><link name=\"link1\"> <visual><geometry><cylinder len"..., 
    srdf="<robot name=\"test_robot\"><group name=\"arm\"><chain base_link=\"base\" tip_link=\"endeff\" /></group><virtual_joint name=\"world_joint\" type=\"fixed\" parent_frame=\"world_frame\" child_link=\"base\" /><group_stat"...) at /home/wxm/tmp_ws/src/exotica/exotica/src/Server.cpp:99
#9  0x00007ffff7b74129 in exotica::Server::getModel (this=0x682030, path="robot_description", 
    urdf="<robot name=\"test_robot\"><link name=\"base\"><visual><geometry><cylinder length=\"0.3\" radius=\"0.2\"/></geometry><origin xyz=\"0 0 0.15\"/></visual></link><link name=\"link1\"> <visual><geometry><cylinder len"..., 
    srdf="<robot name=\"test_robot\"><group name=\"arm\"><chain base_link=\"base\" tip_link=\"endeff\" /></group><virtual_joint name=\"world_joint\" type=\"fixed\" parent_frame=\"world_frame\" child_link=\"base\" /><group_stat"...) at /home/wxm/tmp_ws/src/exotica/exotica/src/Server.cpp:139
#10 0x0000000000415dfe in testRos () at /home/wxm/tmp_ws/src/exotica/exotica/tests/test_initializers.cpp:127
#11 0x0000000000414ed1 in main (argc=1, argv=0x7fffffffd738)
    at /home/wxm/tmp_ws/src/exotica/exotica/tests/test_initializers.cpp:148
(gdb) 
```

PR #99, opened 2017-09-04
Collision and planning scene updates and fixes

### LINKED ISSUE #39: isStateValid returns different results when called with direct or indirect state update
It calls different scenes so returned results are different, cf:

Code:
```cpp
bool is_valid;
is_valid = prob_->getScene()->getCollisionScene()->isStateValid(q);
std::cout << "Direct update without self check: " << is_valid << std::endl;
is_valid = prob_->getScene()->getCollisionScene()->isStateValid(q, true);
std::cout << "Direct update with self check: " << is_valid << std::endl;
prob_->getScene()->getCollisionScene()->update(q);
is_valid = prob_->getScene()->getCollisionScene()->isStateValid();
std::cout << "Two-step update without self check: " << is_valid << std::endl;
prob_->getScene()->getCollisionScene()->update(q);
is_valid = prob_->getScene()->getCollisionScene()->isStateValid(true);
std::cout << "Two-step update with self check: " << is_valid << std::endl;
prob_->getScene()->getCollisionScene()->update(q);
is_valid = prob_->getScene()->getCollisionScene()->isStateValid(true, 0.01);
std::cout << "Two-step update with self check and 1cm dist: " << is_valid << std::endl;
```

Result: (1 = valid, 0 = in collision)
```
Direct update without self check: 0
Direct update with self check: 0
Two-step update without self check: 1
Two-step update with self check: 1
Two-step update with self check and 1cm dist: 1
```

### LINKED ISSUE #100: ROS messages passing through Python wrapper
As discussed via WhatsApp yesterday, we need a way to pass ROS messages through the Python wrapper to Python-wrapped functions. We are currently investigating the use of serialization/deserialization of messages through buffers, however, they currently segfault at the conversion on the C++ side. Additionally, ``StringIO`` isn't available in Python3 anymore: http://python-future.org/compatible_idioms.html#stringio

wxmerkt:
- Adds ``getClosestDistance`` to ``CollisionScene`` with MoveIt collision checking and wraps it for Python (through ``Scene``)
- Removes ``self`` argument from ``isStateValid(q)`` overload as unused (a step towards simplifying #39) - it was unused in all online exotica-based projects
- Unified ``debug_`` and ``visual_debug_`` as the latter couldn't be enabled, ``debug_`` can be set interactively from Python (``debugMode=True``)
- Exposes ``publishScene``
- Adds wrapping for ROS message types (should work for both Python 2 and 3, only tested with Python2 for now) (Resolves #100)
- Add option to update the planning scene from Python via ``moveit_msgs::PlanningSceneWorld``
- Exposes the different overloads for ``isStateValid`` and resolves #39 as cannot be reproduced

### DIFF
--- a/exotica/include/exotica/Scene.h
+++ b/exotica/include/exotica/Scene.h
@@ -128,14 +128,19 @@ namespace exotica
       bool isStateValid(const Eigen::VectorXd &q);
 
       /**
-       * \brief	Get closest distance between robot link and anyother objects
+       * \brief Check for the closest distance in the planning scene including both self- and world-collisions
+       * @return The closest distance - negative values indicate penetration.
+       */
+      double getClosestDistance();
+
+      /**
+       * \brief	Get closest distance between robot link and any other objects
        * @param	link	Robot link
        * @param	self	Indicate if self collision check is required
        * @param	d		Closest distance
        * @param	p1		Closest distance point on the link
        * @param	p2		Closest distance point on the other object
        * @param	norm	Normal vector on robot link
-       * @return Indication of success
        */
       void getRobotDistance(const std::string & link, bool self, double & d,
           Eigen::Vector3d & p1, Eigen::Vector3d & p2, Eigen::Vector3d & norm,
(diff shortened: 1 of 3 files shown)


wxmerkt:
Now ready for review and merge @VladimirIvan

PR #104, opened 2017-09-07
Fix planning scene update from kinematics update

### LINKED ISSUE #103: Kinematica model state update does not trigger planning scene update
The Scene's ``PlanningScene`` joint state (controlled and "uncontrolled") are only updated via ``problem.update(q)``. Updating the model state via ``setModelState()`` or ``setModelStateMap()`` does not update the planning scene states.

wxmerkt:
Includes:

- ``setModelState`` can now be called with subsets of the full state vector when passing in a map - the check was not required and often we get updates from different nodes so won't have the full vector available
- ``setModelState`` has been fixed to trigger recomputation of FK and Jacobian as well as publish debug frames
- Added ``update(joint_name, value)`` method to ``CollisionScene`` to update single joints by name
- Fixed ``Scene::setModelState`` to also update the collision scene, resolves #103

### DIFF
--- a/exotica/include/exotica/Scene.h
+++ b/exotica/include/exotica/Scene.h
@@ -92,6 +92,14 @@ namespace exotica
       void update(Eigen::VectorXdRefConst x);
 
       /**
+       * @brief      Update an individual joint's value in the planning scene
+       *
+       * @param[in]  joint  The joint
+       * @param[in]  value  The value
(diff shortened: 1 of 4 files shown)


VladimirIvan:
There are a couple more ```{``` on the same line as the function declaration or ```if```.

PR #106, opened 2017-09-07
Allow Scenes to have arbitrary base type/ independent of KinematicTree's base type

### LINKED ISSUE #105: Scene for planning group which does not include floating joint wrongly updates floating joint
**Symptom:** problem.update wrongly updates the floating-base when not part of the state

Here's how to reproduce this:

1. Consider a floating-base robot where the current planning group does *not* include the virtual floating joint, i.e. the state is a set of other joints, but fixed w.r.t. to the world
2. Using scene.update(x) or problem.update(x) will incorrectly update the planning scene internal state as the CollisionScene::update function first updates the floating base (which in Kinematica is correctly stored as floating even though for this problem is _not_) and then the joint values offset by the virtual floating joint DoF.

The cause for this is that the ``BaseType`` of the ``Scene`` is incorrectly defined by the ``BaseType`` of the ``KinematicTree`` and then gets passed to ``CollisionScene``.

**Solution:** The ``Scene``/``CollisionScene`` needs to define their own ``BaseType`` based on the planning group used on instantiation.

wxmerkt:
This fixes #105 and also exposes ``getBaseType`` for both ``KinematicTree`` and ``Scene`` as well as ``getGroupName`` for ``Scene``.

### DIFF
--- a/exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
+++ b/exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
@@ -54,6 +54,8 @@ namespace exotica
         void setRho(const std::string & task_name, const double rho);
         Eigen::VectorXd getGoal(const std::string & task_name);
         double getRho(const std::string & task_name);
+        Eigen::VectorXd getNominalPose();
+        void setNominalPose(Eigen::VectorXdRefConst qNominal_in);
 
         Eigen::VectorXd Rho;
         TaskSpaceVector y;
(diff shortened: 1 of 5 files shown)


wxmerkt:
Merging as this fixed several issues and introduced a correct treatment of the floating base. A follow up PR #109 cleans up the code and moves the logic to KinematicTree as discussed as well as fixes a further bug in the treatment of joint limits for floating base robots with joint groups that do not include the floating base.

Issue #96, opened 2017-09-04
KinematicTree/Kinematica uses hard-coded joint limits for floating base joints


wxmerkt:
Cf. https://github.com/openhumanoids/exotica/blob/master/exotica/src/KinematicTree.cpp#L434-L454

VladimirIvan:
How is this handled in MoveIt? Are these specified somewhere in URDF/SRDF or in a additional YAML file?
It would still be good to have a default hard coded value. Maybe with more round numbers.
There is a method for setting these at runtime. The values have to be specified for OMPL to work with floating joints so we need some way to ensure the joint limits are always present.

wxmerkt:
Turns out there was a bug here as well - the joint limits of the first 3 or 6 controlled joints were overwritten if the model had a floating base but the controlled group did not. Fix is in #109.

wxmerkt:
You can set the floating base joint limits in Python, which is what I resort to automatically. You may notice that the particularly chosen values resemble the planning values on _a certain humanoid_. 

Here are the methods for setting the floating base limits (and all of them, if the floating base is actuated):

https://github.com/ipab-slmc/exotica/blob/d446890f49f7b783bcf0004799578cf9d34fd222/exotica/include/exotica/KinematicTree.h#L162-L164

These are exposed to Python (and I use them heavily).

christian-rauch:
I already figured that these values fit well to a particular humanoid robot :-)

I got apparent wrong results using the `JointLimit` task before and therefore never used it. Just now that I searched the source code, I was able to retrace the results.

Because potential future users of exotica may hit similar issues (get weird results) which are hard to debug (searching the source code) I would basically propose to not limit the floating base joints by default.

Also, I disapprove hard coded values in general :-)

christian-rauch:
I am closing this (for good?) since the joint limits have been set to unbound by default in PR https://github.com/ipab-slmc/exotica/pull/435.

PR #109, opened 2017-09-09
KinematicTree BaseType clean-up and floating base joint limits fix

wxmerkt:
This is a follow up to PR #106 which originally fixed the floating base distinction between the _robot model_ and the _controlled joint group_ as well as several other issues. 

This PR focuses on:

- A clean-up of the ``BaseType`` treatment directly in ``KinematicTree`` and moving the logic for the controlled joint group base type as introduced in #106 from ``Scene`` to ``KinematicTree``
- It introduces a clear distinction between the ``ModelBaseType`` and the ``ControlledBaseType``
- A fix for a bug which wrongly set floating base joint limits overriding the joint limits of the controlled group when there was no floating base as part of the controlled group. This was first highlighted when investigating #96

### DIFF
--- a/exotica/include/exotica/KinematicTree.h
+++ b/exotica/include/exotica/KinematicTree.h
@@ -178,7 +178,7 @@ namespace exotica
             ~KinematicTree() {}
             void Instantiate(std::string JointGroup, robot_model::RobotModelPtr model);
             std::string getRootName();
-            BASE_TYPE getBaseType();
+            BASE_TYPE getModelBaseType();
             std::shared_ptr<KinematicResponse> RequestFrames(const KinematicsRequest& request);
             void Update(Eigen::VectorXdRefConst x);
             void setJointLimits();
--- a/exotica/include/exotica/KinematicTree.h
+++ b/exotica/include/exotica/KinematicTree.h
@@ -179,6 +179,7 @@ namespace exotica
             void Instantiate(std::string JointGroup, robot_model::RobotModelPtr model);
             std::string getRootName();
             BASE_TYPE getModelBaseType();
+            BASE_TYPE getControlledBaseType();
             std::shared_ptr<KinematicResponse> RequestFrames(const KinematicsRequest& request);
             void Update(Eigen::VectorXdRefConst x);
             void setJointLimits();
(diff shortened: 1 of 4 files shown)


exotica/src/KinematicTree.cpp
```diff
@@ -199,6 +199,33 @@ void KinematicTree::BuildTree(const KDL::Tree & RobotKinematics)
...
+                    world_joint + "/trans_x") != ControlledJointsNames.end())
+        ControlledBaseType = BASE_TYPE::FLOATING;
+      // A planar base in Kinematica always contains x, y, theta (3-DoF)
+      else if (std::find(ControlledJointsNames.begin(),
+                         ControlledJointsNames.end(),
+                         world_joint + "/theta") !=
+               ControlledJointsNames.end())
+        ControlledBaseType = BASE_TYPE::PLANAR;
+    }
+
```
VladimirIvan: Do you need to check SRDF here?
If the `ModelBaseType`  is Floating or Plannar, the `root_joint==world_joint`. So you can use ```RootJoint->getName()``` to get its name without using SRDF (line 212).
Then, if the root joint is part of the controlled joints ```ControlledBaseType=ModelBaseType``` (replacing lines 220-226).
You can then also check if the root joint is controlled when you add joints into the `ControlledJointsMap` vector at the end of ```BuildTree()```:
```
for(std::shared_ptr<KinematicElement> Joint : Tree)
    {
        Joint->ControlId = IsControlled(Joint);
        Joint->IsControlled = Joint->ControlId >= 0;
        if(Joint->IsControlled) ControlledJoints[Joint->ControlId] = Joint;
        TreeMap[Joint->Segment.getName()] = Joint;
        ModelJointsMap[Joint->Segment.getJoint().getName()] = Joint;
        if(Joint->IsControlled) 
        {
            ControlledJointsMap[Joint->Segment.getJoint().getName()] = Joint;
            \\ This is the new
            if(Joint->Segment.getJoint().getName()==RootJoint->getName() + "/trans_x") ControlledBaseType=BaseType;
        }
    }
```

exotica/include/exotica/KinematicTree.h
```diff
@@ -219,7 +219,7 @@ namespace exotica
             void ComputeJ(const KinematicFrame& frame, KDL::Jacobian& J);
 
             BASE_TYPE ModelBaseType;
-            BASE_TYPE ControlledBaseType;
+            BASE_TYPE ControlledBaseType = BASE_TYPE::FIXED;
```
VladimirIvan: Does this compile?
```ControlledBaseType = BASE_TYPE::FIXED;``` should be done in the constructor or at initialization.

wxmerkt: It does, both on my machine and in Travis (green)

VladimirIvan: ok. It's just non-const static members.

PR #113, opened 2017-09-11
KinematicTree: clear distinction between root joint and root frame

wxmerkt:
- Renames getRootName to getRootFrameName
- Adds getRootJointName

E.g. needed when trying to use the root joint name to retrieve the name of the floating base DoF names when combining with other libraries.

### DIFF
--- a/exotica/src/Scene.cpp
+++ b/exotica/src/Scene.cpp
@@ -729,9 +729,14 @@ namespace exotica
     return collision_scene_;
   }
 
-  std::string Scene::getRootName()
+  std::string Scene::getRootFrameName()
   {
-    return kinematica_.getRootName();
+    return kinematica_.getRootFrameName();
+  }
+
+  std::string Scene::getRootJointName()
+  {
+    return kinematica_.getRootJointName();
   }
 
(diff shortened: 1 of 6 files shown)


exotica/src/Scene.cpp
```diff
@@ -729,9 +729,14 @@ namespace exotica
     return collision_scene_;
   }
 
-  std::string Scene::getRootName()
+  std::string Scene::getRootFrameName()
```
wxmerkt: @VladimirIvan do we even need the root of the scene exposed given that we can just do ``scene.getSolver().getRootFrameName()``?

VladimirIvan: Not necessarily. I do like to expose scene related properties through the scene, even when they originate in the underlying solver.

Also, this PR may affect #114. I've added ```getRobotRootName``` which returns the robot root link name, instead of the scene root. It may need renaming. Feel free to modify both and merge.

wxmerkt: Okay, thanks. I will rename it to ``getModelRootLinkName`` if that's okay

PR #114, opened 2017-09-11
Minor changes required for the HDRM task map

VladimirIvan:
- ```preupdate``` method gets called once at the beginning of ```solve```. This can be used to run one off jobs on problems and solvers.
- Added ```Rho``` to the ```UnconstrainedTimeIndexedProblemInitializer```.
- Cleaned up example configs.
- Added method to retrieve robot root, e.g. `scene_root='world_frame'`, `robot_root='base'`.
- Moved initialization of N from specific problems to the abstract ```PlanningProblem```.

### DIFF
--- a/exotica/include/exotica/Scene.h
+++ b/exotica/include/exotica/Scene.h
@@ -337,7 +337,7 @@ namespace exotica
       CollisionScene_ptr & getCollisionScene();
       std::string getRootFrameName();
       std::string getRootJointName();
-      std::string getRobotRootName();
+      std::string getModelRootLinkName();
       planning_scene::PlanningScenePtr getPlanningScene();
       exotica::KinematicTree & getSolver();
       robot_model::RobotModelPtr getRobotModel();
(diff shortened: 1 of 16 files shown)


exotica/include/exotica/Scene.h
```diff
@@ -336,6 +336,7 @@ namespace exotica
       void setCollisionScene(const moveit_msgs::PlanningSceneConstPtr & scene);
       CollisionScene_ptr & getCollisionScene();
       std::string getRootName();
+      std::string getRobotRootName();
```
wxmerkt: If you look at #114, I've changed L338 as ``Root`` is not clear and added the ability to retrieve the root joint. So in that line, perhaps rename ``getRobotRootName`` to ``getModelRootLinkName``?

exotica_python/src/Exotica.cpp
```diff
@@ -615,6 +615,13 @@ PYBIND11_MODULE(_pyexotica, module)
       }
       return py::make_tuple(d, p1, p2);
     });
+    scene.def("loadScene", &Scene::loadScene);
```
wxmerkt: These are now duplicated - L571 and L620

VladimirIvan:
'hidrm_collision_task' is now up to date with this

PR #120, opened 2017-09-13
Added task map handling to the SmaplingProblem

### LINKED ISSUE #92: Add task map handling to the SamplingProblem
- Add TaskMaps to the Sampling problem.
- Add thresholding mechanism to turn task error into a binary variable.
- Extend the validity checker to use this binary variable.

VladimirIvan:
- Sampling problem now uses task maps, rho and thresholds.
- ```isValid(x)``` method check for validity by computing `rho(Phi-y)<threshold`. Use negative `rho` to emulate the `>=` operator.
- OMPL solver now calls ```isValid```. If no task maps are defined, behaviour is the same as before.
- Fixed `Distance` task map. ```Update(x)``` was failing due to wrong indexing (```Update(x,J)``` overload was working fine).

Fixes #92


wxmerkt:
Thank you - I'll have a detailed look on Monday.

PR #121, opened 2017-09-19
Compute scene object frames in the kinematic tree

VladimirIvan:
- Saves a copy of the robot model tree after loading/initialization.
- Adds ```KinematicTree::AddElement``` for adding extra frames to the kinematics solver.
- ```KinematicElement``` now contains a ```shapes::Shape``` in anticipation of directly feeding this into the ```CollisionScene```, bypassing moveit when setting up FCL.
- Adds ```Scene::updateSceneFrames``` to populate the kinematic tree with objects and colliding shapes currently in the moveit planning scene.

RViz now displays frames of collision shapes.

### DIFF
--- a/exotica/include/exotica/Scene.h
+++ b/exotica/include/exotica/Scene.h
@@ -350,6 +350,7 @@ namespace exotica
       void setModelState(Eigen::VectorXdRefConst x);
       void setModelState(std::map<std::string, double> x);
       std::string getGroupName();
+      void updateSceneFrames();
 
       BASE_TYPE getBaseType()
       {
--- a/exotica/include/exotica/Scene.h
+++ b/exotica/include/exotica/Scene.h
@@ -350,6 +350,8 @@ namespace exotica
       void setModelState(Eigen::VectorXdRefConst x);
       void setModelState(std::map<std::string, double> x);
       std::string getGroupName();
+
+      /// \brief Updates exotica scene object frames from the MoveIt scene.
       void updateSceneFrames();
 
       BASE_TYPE getBaseType()
--- a/exotica/src/KinematicTree.cpp
+++ b/exotica/src/KinematicTree.cpp
@@ -182,21 +182,22 @@ void KinematicTree::BuildTree(const KDL::Tree & RobotKinematics)
     }
 
     AddElement(RobotKinematics.getRootSegment(), *(Tree.end()-1));
+    ModelTree = Tree;
+
+    UpdateModel();
 
     NumJoints = ModelJointsNames.size();
     NumControlledJoints = ControlledJointsNames.size();
     if (NumControlledJoints < 1) throw_pretty("No update joints specified!");
     ControlledJoints.resize(NumControlledJoints);
-    Root = Tree[0];
-    TreeState = Eigen::VectorXd::Zero(Tree.size());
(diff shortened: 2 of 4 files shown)


exotica/src/KinematicTree.cpp
```diff
@@ -213,9 +215,62 @@ void KinematicTree::BuildTree(const KDL::Tree & RobotKinematics)
...
+void KinematicTree::resetModel()
+{
+    Tree = ModelTree;
+    CollisionTree.clear();
+    UpdateModel();
+}
+
+void KinematicTree::AddElement(const std::string& name, Eigen::Affine3d& transform, const std::string& parent, shapes::ShapeConstPtr shape)
+{
+    std::shared_ptr<KinematicElement> parent_emelent;
```
wxmerkt: typo

exotica/src/KinematicTree.cpp
```diff
@@ -213,9 +215,62 @@ void KinematicTree::BuildTree(const KDL::Tree & RobotKinematics)
...
+    if(parent=="")
+    {
+        parent_emelent = Tree[0];
+    }
+    else
+    {
+        bool found = false;
+        for(auto element : Tree)
```
wxmerkt: ``&`` or ``const &`` missing

exotica/src/Scene.cpp
```diff
@@ -840,9 +845,55 @@ namespace exotica
...
+      planning_scene::PlanningScenePtr ps = collision_scene_->getPlanningScene();
+
+      // Add world objects
+      for(auto& object : *ps->getWorld())
+      {
+          if(object.second->shapes_.size())
+          {
+              // Use the first collision shape as the origin of the object
+              Eigen::Affine3d objTransform = object.second->shape_poses_[0];
+              std::string objectName = object.first;
```
wxmerkt: ``&`` reference?

exotica/include/exotica/Scene.h
```diff
@@ -350,6 +350,7 @@ namespace exotica
       void setModelState(Eigen::VectorXdRefConst x);
       void setModelState(std::map<std::string, double> x);
       std::string getGroupName();
+      void updateSceneFrames();
```
wxmerkt: please add doxygen explaining what this function does

exotica/src/Scene.cpp
VladimirIvan: Where? ```object.first``` is a ```std::string```.

exotica/src/KinematicTree.cpp
VladimirIvan: I changed this but here ```auto=std::shared_ptr<KinematicElement>```. Turning it into ```const std::shared_ptr<KinematicElement>&``` does not make a lot of difference. However, it is then called with ```element->Segment.getName()``` which is valid but a bit confusing, because you are calling a reference with the ```->``` operator.

exotica/src/Scene.cpp
wxmerkt: ``std::string& objectName = object.first``

VladimirIvan: That stuff gets optimized out anyway.

VladimirIvan: Ok. I admit it. It doesn't. I have removed the whole variable.

PR #122, opened 2017-09-19
Attached object handling

VladimirIvan:
- Adds attach/detach methods to the scene (bypassing moveit scene)
- Adds scene collision shapes publishing as a marker - only non robot link collision shapes are published. This is much faster to display in RViz than a planning scene.

Octomap is not being published yet.
Objects attached using the moveit scene are not handled correctly. The exotica scene manages attachments separately (and faster). The moveit attachments will not be necessary once the FCL is loaded directly from exotica scene (coming soon).

This PR depends on #121.


exotica/src/KinematicTree.cpp
```diff
@@ -334,6 +431,27 @@ void KinematicTree::publishFrames()
...
+                mrk.action = visualization_msgs::Marker::ADD;
+                mrk.frame_locked = true;
+                mrk.ns = "CollisionObjects";
+                mrk.color.r=mrk.color.g=mrk.color.b=0.5;
+                mrk.color.a=1.0;
+                mrk.header.frame_id = "exotica/"+Tree[i]->Segment.getName();
+                msg.markers.push_back(mrk);
+            }
+        }
+        shapes_pub_.publish(msg);
```
wxmerkt: wrap in if Server::isRos() and potentially if debug_

exotica/src/Scene.cpp
```diff
@@ -840,9 +845,87 @@ namespace exotica
...
+              kinematica_.AddElement(links[i]->getName()+"_collision_"+std::to_string(j), trans, links[i]->getName(), links[i]->getShapes()[j]);
+          }
+      }
+
+      kinematica_.UpdateModel();
+  }
+
+  void Scene::attachObject(const std::string& name, const std::string& parent)
+  {
+      kinematica_.changeParent(name, parent, KDL::Frame(), false);
```
wxmerkt: I think the two overloads could be realised without code duplication. Just call one from the other passing in KDL::Frame() - in case the latter changes in the future

exotica/src/Scene.cpp
```diff
@@ -840,9 +845,87 @@ namespace exotica
...
+  {
+      if(!hasAttachedObject(name)) throw_pretty("'"<<name<<"' is not attached to the robot!");
+      auto object = attached_objects_.find(name);
+      kinematica_.changeParent(name, "", KDL::Frame(), false);
+      attached_objects_.erase(object);
+  }
+
+  bool Scene::hasAttachedObject(const std::string& name)
+  {
+      if(attached_objects_.find(name)!=attached_objects_.end())
```
wxmerkt: Can't this be reduced to ``return (attached_objects_.find(name)!=attached_objects_.end()`` or ``return (attached_objects_.find(name)!=attached_objects_.end() ? true : false``

exotica/src/Scene.cpp
VladimirIvan: I wanted different behaviour here. First overload attaches object from its current wold position, the second specifies the offset relative to the new parent. I will probably rename the methods to make this clearer.

VladimirIvan:
By python script, do you mean an example?

exotica/src/Scene.cpp
wxmerkt: A doxygen to explain that difference may be fine and sufficient.

wxmerkt:
Yes, an example script that we can run that shows e.g. planning to an object, attach, planning a plan with the object, detach, planning to another pose. Simply between 3 poses where one attach and detach sequence happens so we can inspect visually

PR #123, opened 2017-09-26
MotionSolver: be able to provide initial guess from Python

wxmerkt:
Allows passing an initial guess to an optimisation solver (if supported/implemented) from Python, e.g. for AICO

### DIFF
--- a/exotica/include/exotica/MotionSolver.h
+++ b/exotica/include/exotica/MotionSolver.h
@@ -51,6 +51,7 @@ namespace exotica
         virtual void InstantiateBase(const Initializer& init);
         virtual void specifyProblem(PlanningProblem_ptr pointer);
         virtual void Solve(Eigen::MatrixXd & solution) = 0;
+        virtual void Solve(const std::vector<Eigen::VectorXd>& init, Eigen::MatrixXd & solution);
         PlanningProblem_ptr getProblem() {return problem_;}
         virtual std::string print(std::string prepend);
 
--- a/exotica/include/exotica/MotionSolver.h
+++ b/exotica/include/exotica/MotionSolver.h
@@ -51,7 +51,6 @@ namespace exotica
         virtual void InstantiateBase(const Initializer& init);
         virtual void specifyProblem(PlanningProblem_ptr pointer);
         virtual void Solve(Eigen::MatrixXd & solution) = 0;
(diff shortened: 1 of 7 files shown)


exotica/include/exotica/MotionSolver.h
```diff
@@ -51,6 +51,7 @@ namespace exotica
         virtual void InstantiateBase(const Initializer& init);
         virtual void specifyProblem(PlanningProblem_ptr pointer);
         virtual void Solve(Eigen::MatrixXd & solution) = 0;
+        virtual void Solve(const std::vector<Eigen::VectorXd>& init, Eigen::MatrixXd & solution);
```
VladimirIvan: Can you move the ```std::vector<Eigen::VectorXd> init``` into the ```UnconstrainedTimeIndexedProblem``` and let AICO pick it up from there. The solve method can then stay simple and you can specify your init trajectory inside the problem.

wxmerkt: Yes, it will break the current AICO API though.

wxmerkt: Looking at it closer it will require more changes to AICO and we would need to figure out which way we want it in the future:

a) the initial guess is part of the problem and to be set there - requires updating AICO and then iLevator
b) the initial guess is passed in as an argument, similar to how it's done in AICO and iLeavtor

wxmerkt:
@VladimirIvan PTAM - Moved the ``InitialTrajectory`` to ``UnconstrainedTimeIndexedProblem`` and used it in the AICO solver.

exotica/src/Problems/UnconstrainedTimeIndexedProblem.cpp
```diff
@@ -115,6 +115,25 @@ namespace exotica
...
+      const std::vector<Eigen::VectorXd> q_init_in) {
+    if (q_init_in.size() != T)
+      throw_pretty("Expected initial trajectory of length "
+                   << T << " but got " << q_init_in.size());
+    if (q_init_in[0].rows() != N)
+      throw_pretty("Expected states to have " << N << " rows but got "
+                                              << q_init_in[0].rows());
+
+    InitialTrajectory = q_init_in;
+  }
```
VladimirIvan: Add ```setStartState(q_init_in[0]);```

PR #125, opened 2017-09-26
Replaced collision checking from MoveIt by a direct interface to FCL

VladimirIvan:
- Moved ```CollisionScene``` to a separate header and turned it into an abstract class.
- Moved ```KinematicElement``` into a separate header.
- Moved all FCL dependent code into a plugin extending the ```CollisionScene``` class.
- Implemented a direct interface to FCL using collision objects and pointers to ```KinematicElement```s.
- Added ```SceneInitializer``` property to set the collision scene (selecting a plugin to load).
- Added ```fcl_catkin``` package compiling the FCL 0.6.0 from source (submodue).
- Added ```CollisionSceneFCLLatest``` correctly computing correct distances between shapes using FCL 0.6.0 (insteas of the outdated moveit version).
- Added ```CollisionProxy``` class capturing distances between shapes and ```publishProxies``` method to publish them as markers to Rviz.

When computing distances/proxies:
- Collision normal is the direction from one contact point to the other.
- When the shapes are colliding, centres of shapes are used due to insufficient contact points provided by the FCL.
- Distance computation only works with FCL 0.6.0.


exotica/resources/configs/example_manipulate_ompl.xml
```diff
@@ -10,7 +10,6 @@
 
     <PlanningScene>
       <Scene>
-        <!-- Computing distances is enabled by default, disable for a Sampling problem by setting the PlanningMode to speed up Scene updates -->
```
wxmerkt: I think this comment was helpful for people using the example configs to learn about how the system works.

exotica/src/Scene.cpp
```diff
@@ -31,595 +31,14 @@
...
-  }
-
-  void CollisionScene::update(Eigen::VectorXdRefConst x)
-  {
-    if (joint_index_.size() != x.rows())
-      throw_pretty("Size does not match, need vector size of "
-                   << joint_index_.size() << " but " << x.rows()
-                   << " is provided");
-
-    if (BaseType == BASE_TYPE::FIXED)
```
wxmerkt: Since this was such a major pain to fix, just to make sure: does the new CollisionScene work correctly for a) fixed base, b) floating base, c) mixed base, i.e. model is floating base but planning group is fixed base?

exotica_python/src/Exotica.cpp
```diff
@@ -552,6 +553,19 @@ PYBIND11_MODULE(_pyexotica, module)
     samplingProblem.def("getSpaceDim", &SamplingProblem::getSpaceDim);
     samplingProblem.def("getBounds", &SamplingProblem::getBounds);
 
+    py::class_<CollisionProxy, std::shared_ptr<CollisionProxy>> proxy(module, "Proxy");
+    proxy.def(py::init());
+    proxy.def_readonly("Contact1", &CollisionProxy::contact1);
```
wxmerkt: A bit at unease with the inconsistency of using Uppercase at the start of variables - sometimes it's lower case (most common), then in other places suddenly uppercase...

exotica_python/src/Exotica.cpp
```diff
@@ -572,49 +587,11 @@ PYBIND11_MODULE(_pyexotica, module)
...
-      Eigen::Vector3d p1 = Eigen::Vector3d::Zero(),
-                      p2 = Eigen::Vector3d::Zero();
-      if (computeContactInformation) {
-        instance->getCollisionScene()->getDistance(o1, o2, d, p1, p2, safeDist);
-      } else {
-        instance->getCollisionScene()->getDistance(o1, o2, d, safeDist);
-      }
-      return py::make_tuple(d, p1, p2);
-    });
+    scene.def("isStateValid", [](Scene* instance, bool self) {return instance->getCollisionScene()->isStateValid(self);}, py::arg("self")=true);
```
wxmerkt: I'd like the margin/safe distance back - we used it heavily in the past (both in C++ and in Python).

exotica_python/src/Exotica.cpp
```diff
@@ -572,49 +587,11 @@ PYBIND11_MODULE(_pyexotica, module)
...
-    scene.def("getCollisionWorldLinks", [](Scene* instance) {
-        return instance->getCollisionScene()->getCollisionWorldLinks();
-    });
-    scene.def("getRobotDistance", [](Scene* instance, std::string link, bool self, double safeDist) {
-        double d;
-        Eigen::Vector3d p1, p2, norm, c1, c2;
-        instance->getCollisionScene()->getRobotDistance(link, self, d, p1, p2, norm, c1, c2, safeDist);
-        return py::make_tuple(d, norm);
-    });
-    scene.def("getDistance", [](Scene* instance, std::string o1, std::string o2,
```
wxmerkt: I think we should bring this one back.

wxmerkt:
- PlanningMode is now deprecated

exotations/collision_scene_fcl_latest/src/CollisionSceneFCLLatest.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+        const auto& cache_entry = fcl_cache_.find(object.first);
+        if(cache_entry == fcl_cache_.end())
+        {
+            new_object = constructFclCollisionObject(object.second);
+            fcl_cache_[object.first] = new_object;
+        }
+        else
+        {
+            new_object = cache_entry->second;
```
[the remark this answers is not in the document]
VladimirIvan: Nope, just copying the pointer, not the data.

exotations/solvers/ompl_imp_solver/src/ompl_imp_solver/OMPLBaseStateSpace.cpp
```diff
@@ -70,7 +70,7 @@ namespace exotica
         state, q);
 #endif
 
-    if (!prob_->isValid(q) || !prob_->getScene()->getCollisionScene()->isStateValid(self_collision_, margin_))
+    if (!prob_->isValid(q) || !prob_->getScene()->getCollisionScene()->isStateValid(self_collision_))
```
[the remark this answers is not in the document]
VladimirIvan: The margin affects the broad phase solver. It would require a new broad phase solver implementation to add this. This only reduces the number of required broad phase checks though. The current solver will still return a correct result. Possibly with a bit of extra work.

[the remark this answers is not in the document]
wxmerkt: We should definitely keep this in mind and on the list of required updates as our sampling based motion planning currently relies on this. Alternatively we can add link and world padding which MoveIt supports, not sure how easy "inflation" would be with our own FCL Scene.

exotations/solvers/ompl_solver/src/OMPLBaseSolver.cpp
```diff
@@ -31,8 +31,6 @@ namespace exotica
   {
     ompl::base::PlannerData data(ompl_simple_setup_->getSpaceInformation());
     ompl_simple_setup_->getPlanner()->getPlannerData(data);
-    int cnt = prob_->getScene()->getCollisionScene()->stateCheckCnt_;
```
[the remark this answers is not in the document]
VladimirIvan: This number should equal to number of problem updates. If we require this put in a feature request issue.

exotica/include/exotica/CollisionScene.h
```diff
@@ -0,0 +1,137 @@
...
+       * \brief Destructor
+       */
+    virtual ~CollisionScene() {}
+
+    /**
+       * \brief Check if the whole robot is valid (collision only).
+       * @param self Indicate if self collision check is required.
+       * @return True, if the state is collision free..
+       */
+    virtual bool isStateValid(bool self = true) = 0;
```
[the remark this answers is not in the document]
VladimirIvan: See above

exotica/include/exotica/Scene.h
```diff
@@ -38,289 +38,12 @@
...
-                       double safeDist = 0.01);
-
-      /**
-       * \brief Check if the whole robot is valid (collision only)
-       * @param self  Indicate if self collision check is required
-       * @return True, if the state is collision free.
-       */
-      bool isStateValid(bool self = true, double dist = 0);
-
-      /**
```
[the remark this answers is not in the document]
VladimirIvan: They are in `CollisionScene.h` now. Is any of them missing?

exotica/resources/configs/example_manipulate_ompl.xml
VladimirIvan: That is a deprecated parameter. I removed it from the XML.

exotica/src/Scene.cpp
VladimirIvan: There is no base handling in the collision scene. All of that is in the KinematicTree.

exotica_python/src/Exotica.cpp
VladimirIvan: See above.

exotica_python/src/Exotica.cpp
VladimirIvan: Done. The return is a vector of proxies.

exotations/collision_scene_fcl_latest/src/CollisionSceneFCLLatest.cpp
```diff
@@ -0,0 +1,381 @@
...
+        tf::vectorKDLToEigen(c1, p.contact1);
+        tf::vectorKDLToEigen(c2, p.contact2);
+        tf::vectorKDLToEigen(n1, p.normal1);
+        tf::vectorKDLToEigen(n2, p.normal2);
+    }
+    data->Distance = std::min(data->Distance, p.distance);
+    data->Proxies.push_back(p);
+}
+
+bool CollisionSceneFCLLatest::collisionCallbackDistance(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void* data, double& dist)
```
wxmerkt: ``dist`` is now being passed in as a reference but not assigned a value/used as part of the function - computeDistance changes the values in ``data_`` but ``dist`` does not get assigned a value.

exotica/src/Scene.cpp
```diff
@@ -31,595 +31,14 @@
...
-    }
-    else if (BaseType == BASE_TYPE::FLOATING)
-    {
-      ps_->getCurrentStateNonConst().setVariablePosition(
-          world_joint_ + "/trans_x", x(0));
-      ps_->getCurrentStateNonConst().setVariablePosition(
-          world_joint_ + "/trans_y", x(1));
-      ps_->getCurrentStateNonConst().setVariablePosition(
-          world_joint_ + "/trans_z", x(2));
-      KDL::Rotation rot = KDL::Rotation::EulerZYX(x(3), x(4), x(5));
```
wxmerkt: This is why never noticed the ZYX <> RPY bug in Kinematica in practice - Kinematica was never used for floating base and the MoveIt planning scene for sampling based planning was correct due to this.

Cf. #138

PR #135, opened 2017-09-30
Kinetic and 16.04 compatibility

VladimirIvan:
- Moved `fcl_catkin` to `ipab-slmc`.
- Moved and renamed all FCL files within `fcl_catkin` to avoid clashes with system installed FCL.
- Removed unused ```fcl2Eigen```
- Added ````ROS_KINETIC``` flag to `collision_scene_fcl` to deal with API incompatibility introduced in 0.5.0 (`boost::shared_ptr` was changed to `std::shared_ptr`).
- Fixed Kinetic detection in OMPL solver.
- Fixed exported include directories in packages where header files were generated directly into devel.
- Fixed Python string type detection when creating initializers.

`collision_cene_fcl_latest` segfaults on 16.04 when computing collision distance. The unit tests provided by FCL all succeed on the same system though. All other features work fine.

### DIFF
--- a/exotations/collision_scene_fcl_latest/CMakeLists.txt
+++ b/exotations/collision_scene_fcl_latest/CMakeLists.txt
@@ -4,15 +4,14 @@ project(collision_scene_fcl_latest)
 add_compile_options(-std=c++11)
 
 find_package(catkin REQUIRED COMPONENTS
-  exotica
-  geometric_shapes
   fcl_catkin
+  geometric_shapes
 )
 
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES collision_scene_fcl_latest
-  CATKIN_DEPENDS exotica geometric_shapes fcl_catkin
+  CATKIN_DEPENDS fcl_catkin geometric_shapes
 )
 
 include_directories(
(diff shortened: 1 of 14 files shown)


exotations/collision_scene_fcl_latest/CMakeLists.txt
```diff
@@ -3,21 +3,28 @@ project(collision_scene_fcl_latest)
 
 add_compile_options(-std=c++11)
 
+
 find_package(catkin REQUIRED COMPONENTS
```
wxmerkt: Why does this have to be a two-step find_package catkin?

wxmerkt: Actually, I suspect this is to get a different order of multiple FCL include dirs in the array of include dirs, because the catkin-found one is different from the one that comes through exotica, correct? Perhaps add an explanatory comment

VladimirIvan: I needed to compile against the exotica headers but not link against its libraries. Only the last `find_package` is used for linking.
The first one extracts the exotica header paths and the second one sets the linked libraries. I didn't want to link against exotica as that would link the the wrong version of FCL. I'm not entirely sure if this solves the problem once the plugin is loaded, but the SO file has a lot fewer dependencies (checked through `ldd`).

PR #137, opened 2017-10-01
Fix root inertial for floating-base robots and Center of Mass computation bug, fixes trans_x not existing, fixes floating base rotation

wxmerkt:
Resolves #134 
Resolves #136 
Resolves #138 

- Fixes extraction and assignment of root inertials, i.e. the mass and Center of Gravity of the root link which KDL neglects
- Fixes advertising & publishing of debug markers for the CoM task
- Fixes wrong center of mass calculation due to wrong relative frame/transform - verified against Drake for an all-zero configuration -- bug got introduced in https://github.com/ipab-slmc/exotica/commit/e89564c1647f742c236520bc604bb729630dd945
- Fixes wrong RPY convention and updating - now YPR with rotating axis

PR #139, opened 2017-10-02
Add floating-base unit test for CoM (Valkyrie)

wxmerkt:
- Based on top of #137 
- Adds Valkyrie URDF and SRDF, without any meshes (so won't be able to visualise)
- Adds a unit test testing the CoM of Valkyrie against values obtained from Drake's RigidBodyManipulator (Matlab)


VladimirIvan:
This is also good to go when travis finishes.

Issue #131, opened 2017-09-29
Decide on standard clang-format style


wxmerkt:
E.g. Google

VladimirIvan:
On top of that, we also need to revisit the naming conventions:
- Static/member/local/argument variable names.
- Consistent use of underscores and capitalization.
- Class, struct, enum naming.
- Static/member and public/private function names.

VladimirIvan:
So here are some proposals:
Formatting:
- Google style - available as CLang format but the curly bracket placement bothers me.
- ROS style - not officially supported and [untested CLang](https://github.com/davetcoleman/roscpp_code_format/blob/master/.clang-format), most of the current code uses this format now, I prefer this format.

Naming convention:
- [ROS convention](http://wiki.ros.org/CppStyleGuide).
- [ETH convention](https://github.com/ethz-asl/programming_guidelines/wiki/Cpp-Coding-Style-Guidelines) more detailed, couple of minor differences in capitalization (compared to ROS).

wxmerkt:
I'd opt for clang-format-3.8 Google style. ROS style is not uniform and changing, the CLang isn't really tested. clang-format (and the different versions) is available on all platforms.

Regarding naming convention, no major preference but I'd go with a big standard as well. Not to be 100% strict with that but good to keep as a guideline.

VladimirIvan:
Formatting fixed in #141.

Todo: fix naming the convention.

wxmerkt:
**C++**

- [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Naming)
- [Drake C++ Style Guide](https://drake.mit.edu/styleguide/cppguide.html#General_Naming_Rules)
- [Drake Robotics Terminology and Notation](http://drake.mit.edu/doxygen_cxx/group__multibody__notation.html)

**Python**

- [PEP 8](https://www.python.org/dev/peps/pep-0008/#naming-conventions)

VladimirIvan:
I went through the google style guide and PEP8 and further specified the style where there were multiple options. I also added suggestions for package and directory naming convention based off ROS standards.

# Packages, files and directories
 - Package names in lower case with underscores and exotica prefix:
```
exotica
exotica_python
exotica_examples
exotica_core_taskmaps
exotica_aico_solver
exotica_ik_solver
exotica_collision_scene_fcl
...
```
 - All solver packages with `_solver` suffix.
 - All task map packages with `_taskmaps` suffix.
[...]

PR #141, opened 2017-10-03
Automatic reformatting using clang-format-3.8 and style=google

wxmerkt:
Proposal to address #131 

- Command used: ``find -name '*.cpp' -o -name '*.h' | xargs clang-format-3.8 -style=google -i``, committed under neutral author Exotica Dev Team
- On top of a hotfix required for it to work: https://github.com/ipab-slmc/exotica/commit/18254951cd523eeb7c9e453b5ea690df36a1bcf6
- Will add an automatic pre-commit-hook that runs ``clang-format-3.8`` before any commit in a follow-up PR (i.e. with clang-format install no further user action required)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+# -*- yaml -*-
+BasedOnStyle: Google
+Language: Cpp
+IndentWidth: 4
+ColumnLimit: 0 # do not enforce a column limit, users are advised to make breaks as they see fit to ensure readability/logic
+BreakBeforeBraces: Allman
+AccessModifierOffset: -4
--- a/examples/exotica_examples/src/core.cpp
+++ b/examples/exotica_examples/src/core.cpp
@@ -36,6 +36,6 @@ using namespace exotica;
 
 int main(int argc, char **argv)
 {
-    ROS_INFO_STREAM("Exotica version: "<<exotica::Version);
+    ROS_INFO_STREAM("Exotica version: " << exotica::Version);
     Setup::printSupportedClasses();
 }
--- a/examples/exotica_examples/src/generic.cpp
+++ b/examples/exotica_examples/src/generic.cpp
@@ -39,36 +39,34 @@ void run()
     Server::InitRos(std::shared_ptr<ros::NodeHandle>(new ros::NodeHandle("~")));
 
     // Scene using joint group 'arm'
-    Initializer scene("Scene",{
-                          {"Name",std::string("MyScene")},
-                          {"JointGroup",std::string("arm")},
-                          {"URDF",std::string("{exotica}/resources/robots/lwr_simplified.urdf")},
-                          {"SRDF",std::string("{exotica}/resources/robots/lwr_simplified.srdf")}});
+    Initializer scene("Scene", {{"Name", std::string("MyScene")},
+                                {"JointGroup", std::string("arm")},
+                                {"URDF", std::string("{exotica}/resources/robots/lwr_simplified.urdf")},
+                                {"SRDF", std::string("{exotica}/resources/robots/lwr_simplified.srdf")}});
     // End-effector task map with two position frames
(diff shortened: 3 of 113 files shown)


wxmerkt:
@VladimirIvan I went through the code and the line-breaking seems okay. Where it's really annoying or we prefer not to wrap or do our own indentation for whatever reason we can either end that line with ``// clang-format off`` or encapsulate a block by ``// clang-format off`` and ``// clang-format on``. I've added an example of where this made a difference to keep the original indentation (exotica_examples/src/generic.cpp). Overall, I think the formatter gives better readability with the automatic wrapping even if it's sometimes a bit too aggressive for some lines, where we can then control manually. 80 seems like a fine limit for having 2 documents next to each other on a landscape widescreen or a single file on a portrait-oriented screen. Additionally, it makes a lot of sense when comparing changes e.g. here on GitHub during reviews where we have two files side by side. Happy to revert/fix any blocks in the changes that it was too aggressive, but generally probably better if we stick to the standard.

VladimirIvan:
Here are a couple of examples where the 80 character line break is just plain silly:
```
 static std::shared_ptr<fcl::CollisionObjectd> constructFclCollisionObject(
     std::shared_ptr<KinematicElement> element);
 static void checkCollision(fcl::CollisionObjectd* o1,
                            fcl::CollisionObjectd* o2, CollisionData* data);
 static void computeDistance(fcl::CollisionObjectd* o1,
                             fcl::CollisionObjectd* o2, DistanceData* data);
 if (info != 0) {
   throw_pretty("Can't invert matrix. Cholesky decomposition failed: " << info
                                                                        << "\n"
                                                                        << A_);
 }
```
Adding a formatting exception before each of those is just too much.

I'm having my doubts about the google style as well:
```
 if (J.cols() < J.rows()) {
   //(Jt*C^-1*J+W)^-1*Jt*C^-1
   if (parameters_.C != 0) {
     Jpinv =
         inverseSymPosDef(J.transpose() * Cinv * J + W) * J.transpose() * Cinv;
   } else {
     Jpinv = inverseSymPosDef(J.transpose() * J + W) * J.transpose();
   }
 } else {
   // W^-1*Jt(J*W^-1*Jt+C)
   Jpinv =
       Winv * J.transpose() * inverseSymPosDef(J * Winv * J.transpose() + C);
 }
```
Having just two spaces for indentation makes it really hard to see where block belong to in longer statements. I literally have to stop reading the code after every line. Below is what the original was intended to look like:
```
if (J.cols() < J.rows()) 
{
    //(Jt*C^-1*J+W)^-1*Jt*C^-1
    if (parameters_.C != 0) 
    {
        Jpinv = inverseSymPosDef(J.transpose() * Cinv * J + W) * J.transpose() * Cinv;
    } 
    else 
    {
        Jpinv = inverseSymPosDef(J.transpose() * J + W) * J.transpose();
    }
} 
else 
{
    // W^-1*Jt(J*W^-1*Jt+C)
    Jpinv = Winv * J.transpose() * inverseSymPosDef(J * Winv * J.transpose() + C);
}
```

I'm no longer convinced google style is the right way to go at all. I'll keep looking for a solution.

VladimirIvan:
Actually, Google with these three changes is fine:
```
IndentWidth: 4
ColumnLimit: 500
BreakBeforeBraces: Allman
```

wxmerkt:
Added config file and updated this PR. The command can be run from the root of the repository with ``find -name '*.cpp' -o -name '*.h' | xargs clang-format-3.8 -style=.clang-format -i``.

**I still think we should think about a sensible column limit. Our merge tools - both GitHub and meld - are pretty poor with inline changes over long lines so reviewing and merging changes will become more difficult with this high of a limit. Open for any suggestions but the shorter the better for both GitHub reviews as well as manual merges.**

PR #142, opened 2017-10-06
Trajectory generators

VladimirIvan:
- Added the ```Trajectory``` class, wrapping the ```KDL::Trajectory_Composite```, ```KDL::Path_line``` and spline interpolation. This was necessary since ```KDL::Trajectory``` read/write methods are broken. SE(3) frame along the trajectory can be queried at time `t`.
- Trajectory can be stored and loaded from a string or generated from an ```Eigen::MatrixXd``` matrix of Nx4 (t+xyz), Nx7 (t+xyzrpy) or Nx8 (t+xyzxyzw) elements for N trajectory via-points.
- Added trajectory generators to the scene. These are pairs of ```KinematicElement``` and ```Trajectory```. Every time a generator is updated, the local frame of the element is chnged based on the pose obtained from the trajectory at time `t`.
- Added ```KinematicElement::getPose``` replacing ```KDL::Segement::pose```. This method returns a frame generated from a trajectory if the generator mode is enabled for the element.
- ```Scene::Update``` now takes `t` as an optional parameter (t=0 by default). The generators are automatically updated to time `t`. If no generators are defined no updates are performed.
- AICO now calls update with the correct argument `t`.
- Python trajectory publish method also correctly updates the time `t`.
- Python AICO example has been updated to use this feature using a trajectory of figure 8 stored in a file.
- Collision scene, additional frames and trajectory generators can now be specified in the ```SceneInitializer```.
- Simplified loading in couple of python examples.

TODO: wrap Trajectory in python.


exotica/include/exotica/Scene.h
```diff
@@ -118,11 +127,13 @@ namespace exotica
...
       }
 
+      void updateTrajectoryGenerators(double t = 0);
+
       void publishScene();
       void publishProxies(const std::vector<CollisionProxy>& proxies);
       visualization_msgs::Marker proxyToMarker(const std::vector<CollisionProxy>& proxies, const std::string& frame);
-      void loadScene(const std::string& scene);
-      void loadSceneFile(const std::string& file_name);
+      void loadScene(const std::string& scene, bool updaetCollisionScene = true);
```
wxmerkt: typo in ``update``

exotica/include/exotica/Scene.h
```diff
@@ -118,11 +127,13 @@ namespace exotica
...
+      void loadSceneFile(const std::string& file_name, bool updaetCollisionScene = true);
```
wxmerkt: typo

exotica/resources/configs/aico_solver_demo_eight.xml
```diff
@@ -17,22 +17,29 @@
...
+          <Trajectory Link="Target" File="{exotica}/resources/scenes/figure_eight.traj" />
+        </Trajectories>
       </Scene>
     </PlanningScene>
 
     <Maps>
       <EffFrame Name="Frame">
         <EndEffector>
-            <Frame Link="lwr_arm_6_link" LinkOffset="0 0 0.2 0.7071067811865476 -4.3297802811774664e-17  0.7071067811865475 4.3297802811774664e-17" BaseOffset="0.6 -0.1 0.5 0 0 0 1" />
+            <Frame Link="lwr_arm_6_link" LinkOffset="0 0 0.2 0.7071067811865476 -4.3297802811774664e-17  0.7071067811865475 4.3297802811774664e-17" Base="Target" />
```
wxmerkt: the e-17 values could likely be replaced with 0

exotica/src/Scene.cpp
```diff
@@ -283,20 +325,20 @@ namespace exotica
       return group->getName();
   }
 
-  void Scene::loadScene(const std::string& scene)
+  void Scene::loadScene(const std::string& scene, bool updaetCollisionScene)
```
wxmerkt: typo

exotica/src/Scene.cpp
```diff
@@ -283,20 +325,20 @@ namespace exotica
...
   {
       std::stringstream ss(scene);
       ps_->loadGeometryFromStream(ss);
       updateSceneFrames();
-      collision_scene_->updateCollisionObjects(kinematica_.getCollisionTreeMap());
+      if(updaetCollisionScene) collision_scene_->updateCollisionObjects(kinematica_.getCollisionTreeMap());
```
wxmerkt: typo

exotica/src/Scene.cpp
```diff
@@ -283,20 +325,20 @@ namespace exotica
...
   }
 
-  void Scene::loadSceneFile(const std::string& file_name)
+  void Scene::loadSceneFile(const std::string& file_name, bool updaetCollisionScene)
```
wxmerkt: typo

exotica/src/Scene.cpp
```diff
@@ -283,20 +325,20 @@ namespace exotica
...
   {
       std::ifstream ss(parsePath(file_name));
       ps_->loadGeometryFromStream(ss);
       updateSceneFrames();
-      collision_scene_->updateCollisionObjects(kinematica_.getCollisionTreeMap());
+      if(updaetCollisionScene) collision_scene_->updateCollisionObjects(kinematica_.getCollisionTreeMap());
```
wxmerkt: typo

exotica/src/Scene.cpp
```diff
@@ -380,6 +444,41 @@ namespace exotica
...
+      if(it==tree.end()) throw_pretty("Can't find link '"<<link<<"'!");
+      if(traj->getDuration()==0.0) throw_pretty("The trajectory is empty!");
+      trajectory_generators_[link] = std::pair<std::shared_ptr<KinematicElement>, std::shared_ptr<Trajectory>>(it->second, traj);
+      it->second->IsTrajectoryGenerated = true;
+  }
+
+  std::shared_ptr<Trajectory> Scene::getTrajectory(const std::string& link)
+  {
+      const auto& it = trajectory_generators_.find(link);
+      if(it==trajectory_generators_.end()) throw_pretty("No trajectory genrator defined for link '"<<link<<"'!");
```
wxmerkt: typo

exotica/src/Scene.cpp
```diff
@@ -380,6 +444,41 @@ namespace exotica
...
+      return it->second.second;
+  }
+
+  void Scene::removeTrajectory(const std::string& link)
+  {
+      const auto& it = trajectory_generators_.find(link);
+      if(it==trajectory_generators_.end()) throw_pretty("No trajectory genrator defined for link '"<<link<<"'!");
+      it->second.first->IsTrajectoryGenerated = false;
+      trajectory_generators_.erase(it);
```
wxmerkt: should this also delete the link?

exotica/src/Scene.cpp
```diff
@@ -380,6 +444,41 @@ namespace exotica
...
+  {
+      const auto& it = trajectory_generators_.find(link);
+      if(it==trajectory_generators_.end()) throw_pretty("No trajectory genrator defined for link '"<<link<<"'!");
+      return it->second.second;
+  }
+
+  void Scene::removeTrajectory(const std::string& link)
+  {
+      const auto& it = trajectory_generators_.find(link);
+      if(it==trajectory_generators_.end()) throw_pretty("No trajectory genrator defined for link '"<<link<<"'!");
```
wxmerkt: typo

exotica_python/scripts/example_aico.py
```diff
@@ -1,26 +1,17 @@
 #!/usr/bin/env python
```
wxmerkt: I think we should offer the old and new example instead of overwriting - i.e. an example_aico.py as currently is and then an exampe_aico_trajectory_generator.py with the new.

exotica/src/Scene.cpp
VladimirIvan: Could do. It will witch to the original KDL::Segment offset after this in the current implementation. I don't expect this to be used at all though. The scene is supposed to be set up once and stay that way until its destroyed. I added removing just to keep the trajectories editable.

Issue #156, opened 2017-10-16
Collision Proxies returned for pairs that are disabled in ACM


wxmerkt:
Is that by design?

VladimirIvan:
The overload of ```getCollisionDistance``` that takes the names of the links does not check for ACM because you specifically requested the links. You can add it as a optional feature if you need it.
The other one does check for ACM.

wxmerkt:
Verified this is a bug:

![image](https://user-images.githubusercontent.com/1664508/31671609-3dcda290-b352-11e7-8ba6-080b7f74f4f4.png)

wxmerkt:
After switching to INDEP per #159 we get some really wild CollisionProxies in an otherwise empty scene:

![image](https://user-images.githubusercontent.com/1664508/31679728-ed63e856-b369-11e7-98da-4fa0789f13c9.png)

PR #159, opened 2017-10-17
Fix collision checking: change to INDEP, update object transforms when in lazy checking mode

### LINKED ISSUE #155: Collision Checking with Meshes Broken
Just noticed that with the new collision scene OMPL is broken for a scene with meshes (the same I have been using for weeks) - i.e. RRT plans a straight line through the mesh as if it wasn't there. This essentially breaks all our collision checking against meshes.

Even for the normal IK scene when I visually cannot see any collision in neither the planning scene nor the published collision shapes it returns 943 collision proxies with ~650 being closer than a centimetre.

This is a more general case of #153 - not only the points are wrong, the entire collision check is broken (both for old and new FCL).

wxmerkt:
- Passes information on whether the collision scene is updated lazily to the CollisionScene
- Uses that information to force an object transform update when queried and it has not been externally updated
- This is e.g. of relevance when using multiple task maps that query the CollisionScene - in that case only one update would be required for multiple queries (and that update would be from AlwaysUpdateCollisionScene, i.e. forced from Scene)
- Switches to INDEP for collision queries in addition to distance queries as per recommendation/tests from Vlad
- Adds method to Scene to allow querying whether the Scene is updated lazily (also used in another branch). This may be replaced with checking it directly from CollisionScene in the future
- The tests for collision checking are *not* enabled yet due to missing meshes. We should add tests e.g. using the ur_description with meshes that can be apt-get install'ed
- Relates to #158 

- Resolves #155, #156

### DIFF
--- a/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
+++ b/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
@@ -222,6 +222,8 @@ bool CollisionSceneFCL::collisionCallback(fcl::CollisionObject* o1, fcl::Collisi
 
 bool CollisionSceneFCL::isStateValid(bool self, double safe_distance)
 {
+    if (!alwaysExternallyUpdatedCollisionScene_) updateCollisionObjectTransforms();
+
     std::shared_ptr<fcl::BroadPhaseCollisionManager> manager(new fcl::DynamicAABBTreeCollisionManager());
     manager->registerObjects(fcl_objects_);
     CollisionData data(this);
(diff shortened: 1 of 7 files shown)


wxmerkt:
CollisionProxies are currently broken both in master as well as after this PR (but less so for meshes). Switching to INDEP and adding 5c0c120 helps in fixing them somewhat, but even the collisionDistance test in exotica_python does not correctly work in master anymore.

wxmerkt:
Now fixed it for both primitive shapes and meshes - the issue was that FCL returns contact points for meshes in world frame while for primitive shapes in local frame. The collision distance demo with the primitive LWR still runs while my Valkyrie results look more realistic (still a few spurious proxies but the main ones work).

We will likely still be subject to https://github.com/flexible-collision-library/fcl/issues/221

Issue #165, opened 2017-10-18
Distinguish robot and world safe distances in collision check [Investigate poor OMPL Performance]


wxmerkt:
Model State: 39 DoF, Controlled Joints: 10

RRT-Connect for the well-known scenario:
```
Info:    Exotica_Exotica_RRTConnect: Created 713 states (702 start + 11 goal)
Info:    Solution found in 275.864332 seconds
```

It's complex, sure, but 713 states in 275s is... bad.

wxmerkt:
Found the issue! It's the distinction between robot and environment safe distance I've been talking about - disabling the SafeDistance:

```
Info:    Exotica_Exotica_RRTConnect: Starting planning with 1 states already in datastructure
Info:    Exotica_Exotica_RRTConnect: Created 381 states (372 start + 9 goal)
Info:    Solution found in 0.941445 seconds
```

Will work on that one and push a fix. I.e. the change in ``CollisionCheck`` ``SafeDistance`` behaviour between the original ``isStateValid`` and the new ``isStateValid`` caused a disproportionate number of samples to be invalid (torso and pelvis too close).

wxmerkt:
Additionally, we should look into a different way on how to enforce safe distances - distance queries are very, very expensive. Perhaps padding of the links and maintaining two different trees for self- and external collision checking.

wxmerkt:
The ``reinterpret_cast`` is so expensive that disabling the checks between robot links has no benefit:

```python
%timeit ompl_scene.isStateValid(True, 0.0)
%timeit ompl_scene.isStateValid(False, 0.0)
%timeit ompl_scene.isStateValid(True, 0.03)
%timeit ompl_scene.isStateValid(False, 0.03)
```

#### Current master (does not disable safe distances between robot links)
```
123 µs ± 188 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
28.8 µs ± 92 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
**81.2 µs ± 152 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)**
29 µs ± 546 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

#### Disabling safe distance for robot links (requires 2 reinterpret_cast)
```
124 µs ± 222 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
[...]

PR #167, opened 2017-10-19
Change CollisionScene to shared_ptr

wxmerkt:
As required for the smooth collision cost function, and also to disentangle the performance drop from the safe distance issue in #165. The move definitely results in some slight performance drop, as below. But nothing what we've seen when disabling safe distances for robots.

# Master

121 µs ± 270 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
26.6 µs ± 31.1 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
80.4 µs ± 164 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
26.6 µs ± 86.5 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

# Shared Ptr

123 µs ± 433 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
28.7 µs ± 76.2 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
80.6 µs ± 66.6 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
28.5 µs ± 60.2 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

126 µs ± 2.62 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
29.8 µs ± 939 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
80.9 µs ± 168 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
29.1 µs ± 58.9 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

### DIFF
--- a/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
+++ b/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
@@ -57,7 +57,7 @@ void CollisionSceneFCL::updateCollisionObjects(const std::map<std::string, std::
 {
     kinematic_elements_ = MapToVec(objects);
     fcl_objects_.resize(objects.size());
-    int i = 0;
+    long i = 0;
     for (const auto& object : objects)
     {
         std::shared_ptr<fcl::CollisionObject> new_object;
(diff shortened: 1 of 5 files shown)


exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
```diff
@@ -162,17 +166,17 @@ std::shared_ptr<fcl::CollisionObject> CollisionSceneFCL::constructFclCollisionOb
             throw_pretty("This shape type (" << ((int)shape->type) << ") is not supported using FCL yet");
     }
     geometry->computeLocalAABB();
-    geometry->setUserData(reinterpret_cast<void*>(element.get()));
+    geometry->setUserData(reinterpret_cast<void*>(i));
     std::shared_ptr<fcl::CollisionObject> ret(new fcl::CollisionObject(geometry));
-    ret->setUserData(reinterpret_cast<void*>(element.get()));
+    ret->setUserData(reinterpret_cast<void*>(i));
```
VladimirIvan: I think you may be casting ```int i``` into ```void *``` (64bit long on 64bit architecture) and then later back into ```long```.
Change ```int i``` to ```long i```. I don't think this is causing any problems though.

wxmerkt: Agreed - this is a remnant of when we changed the latter cast to long to avoid compile errors. Will change and update.

PR #169, opened 2017-10-19
Add link/object scale and padding to CollisionScene

### LINKED ISSUE #165: Distinguish robot and world safe distances in collision check [Investigate poor OMPL Performance]

wxmerkt:
This is an alternative to disabling safe distance for robot links which has proven to be much, much more efficient.

E.g. on Valkyrie:

```
Info:    Exotica_Exotica_RRTConnect: Created 1340 states (1316 start + 24 goal)
Info:    Solution found in 6.344722 seconds
```
I.e. one valid state every 4.8ms vs 387ms in #165 

Resolves #165

### DIFF
--- a/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
+++ b/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
@@ -96,6 +96,27 @@ std::shared_ptr<fcl::CollisionObject> CollisionSceneFCL::constructFclCollisionOb
     // Maybe use cache here?
 
     shapes::ShapeConstPtr shape = element->Shape;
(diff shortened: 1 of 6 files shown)


wxmerkt:
Good catch - in that case on the next ``updateCollisionObjects`` it would use the padding of a robot link rather than the world link padding.

PR #171, opened 2017-10-20
Reduce computations by only updating maps if required

### LINKED ISSUE #170: Do not update TaskMaps where Rho=0 for the timestep to safe computation
This can reduce problem update time quite considerably if an expensive TaskMap is not being taken into account during a timestep, i.e. its associated ``Rho`` is 0.

Proposed solution: If a TaskMap's Rho at the current timestep is 0, set Phi and Jacobian to 0, else compute as currently.

wxmerkt:
Resolves #170

### DIFF
--- a/exotica/src/Problems/SamplingProblem.cpp
+++ b/exotica/src/Problems/SamplingProblem.cpp
@@ -125,13 +125,7 @@ bool SamplingProblem::isValid(Eigen::VectorXdRefConst x)
     for (int i = 0; i < NumTasks; i++)
     {
         if (Rho(i) != 0)
-        {
             Tasks[i]->update(x, Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length));
-        }
-        else
-        {
-            Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length).setZero();
-        }
     }
     return ((S * (Phi - y) - threshold_).array() < 0.0).all();
 }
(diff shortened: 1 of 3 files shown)


exotica/src/Problems/SamplingProblem.cpp
```diff
@@ -124,7 +124,14 @@ bool SamplingProblem::isValid(Eigen::VectorXdRefConst x)
...
     for (int i = 0; i < NumTasks; i++)
     {
-        Tasks[i]->update(x, Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length));
+        if (Rho(i) != 0)
+        {
+            Tasks[i]->update(x, Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length));
+        }
+        else
+        {
+            Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length).setZero();
```
VladimirIvan: This line will cause line 136 below to fail when the ```TaskSpaceVector``` Phi contains any quaternion rotations.
Setting the data to zero will set ```xyzw``` to zero creating an invalid rotation.
There is no easy way to directly set the sub segment to zero rotation, except:
Somewhere in the initialization, after the Phi was constructed, create a copy of it and set the copy to zero:
```
TaskSpaceVector ZeroTask = Phi;
ZeroTask.setZero()
```
This will set all the rotations to their correct zero values.
Then in the line above do:
```
Phi.data.segment(Tasks[i]->Start, Tasks[i]->Length) = ZeroTask.data.segment(Tasks[i]->Start, Tasks[i]->Length);
```
This appears in every problem type below as well.

The rest of the PR looks good.

VladimirIvan: Actually, an even cheaper option is to completely skip this line. It will get multiplied by a zero so any valid value can stay in Phi.

wxmerkt: Good point, thank you. I will update accordingly.

PR #182, opened 2017-11-02
KinematicTree: Fix computation of Jacobians for frames with offsets

### LINKED ISSUE #181: Relative point in Jacobian not respected - same as "0,0,0" relative point
When requesting Jacobians for offset points in a link, the returned Jacobian is the same as if the offset were "0,0,0".

wxmerkt:
The issue in #181 was caused by the temporary frames for the desired offsets (in-link/in-frame positions) not being created when computing Jacobians. This PR adds an explicit update of the internal temporary offset frames when computing Jacobians.

Fixes #181

### DIFF
--- a/exotica/include/exotica/KinematicTree.h
+++ b/exotica/include/exotica/KinematicTree.h
@@ -99,10 +99,10 @@ public:
 
 struct KinematicFrame
 {
-    std::shared_ptr<KinematicElement> FrameA = nullptr;
-    KDL::Frame FrameAOffset = KDL::Frame();
-    std::shared_ptr<KinematicElement> FrameB = nullptr;
-    KDL::Frame FrameBOffset = KDL::Frame();
+    std::shared_ptr<KinematicElement> FrameA;
+    KDL::Frame FrameAOffset;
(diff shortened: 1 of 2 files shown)


exotica/include/exotica/KinematicTree.h
```diff
@@ -99,10 +99,10 @@ class KinematicsRequest
...
 {
-    std::shared_ptr<KinematicElement> FrameA;
-    KDL::Frame FrameAOffset;
-    std::shared_ptr<KinematicElement> FrameB;
-    KDL::Frame FrameBOffset;
+    std::shared_ptr<KinematicElement> FrameA = nullptr;
+    KDL::Frame FrameAOffset = KDL::Frame();
+    std::shared_ptr<KinematicElement> FrameB = nullptr;
+    KDL::Frame FrameBOffset = KDL::Frame();
     KDL::Frame TempAB;
```
VladimirIvan: I believe all these initializations are redundant. Both ```shared_ptr``` and ```KDL::Frame``` will be initialized to those values by default.

PR #185, opened 2017-11-03
Optimisation problem improvements

VladimirIvan:
- Seths RViz to 30FPS
- Resets Phi and Jacobian to zero before problem update (fixes NaN issues).
- Fixed pre-update callback in time-indexed problem. The call was not propagated correctly.
- Changed the sign on ```ydiff```. This didn't affect IK and AICO because they always computed the squared error (removing the sign).
- Moved setting up the S matrix into the problem(s) (diagonal matrix made out of Rho for each task).
- Added a scalar cost and scalar cost Jacobian methods for end-pose and time-indexed problems.

### DIFF
--- a/examples/exotica_examples/resources/configs/ik_solver_demo.xml
+++ b/examples/exotica_examples/resources/configs/ik_solver_demo.xml
@@ -26,6 +26,8 @@
         </EndEffector>
       </EffFrame>
     </Maps>
+    <StartState>0 0 0 0 0 0 0</StartState>
+    <NominalState>0 0 0 0 0 0 0</NominalState>
     <W> 7 6 5 4 3 2 1 </W>
   </UnconstrainedEndPoseProblem>
 
--- a/exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
+++ b/exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
@@ -56,13 +56,19 @@ public:
     double getRho(const std::string& task_name);
     Eigen::VectorXd getNominalPose();
     void setNominalPose(Eigen::VectorXdRefConst qNominal_in);
+    virtual void preupdate();
+
+    double getScalarCost();
+    Eigen::VectorXd getScalarJacobian();
 
     Eigen::VectorXd Rho;
     TaskSpaceVector y;
+    Eigen::VectorXd ydiff;
     Eigen::MatrixXd W;
     TaskSpaceVector Phi;
     Eigen::MatrixXd J;
     Eigen::VectorXd qNominal;
+    Eigen::MatrixXd S;
 
     int PhiN;
     int JN;
--- a/exotica/src/Problems/UnconstrainedEndPoseProblem.cpp
+++ b/exotica/src/Problems/UnconstrainedEndPoseProblem.cpp
@@ -114,6 +114,8 @@ Eigen::VectorXd UnconstrainedEndPoseProblem::getScalarJacobian()
 void UnconstrainedEndPoseProblem::Update(Eigen::VectorXdRefConst x)
 {
     scene_->Update(x);
+    Phi.setZero(PhiN);
+    J.setZero();
     for (int i = 0; i < NumTasks; i++)
     {
         if (Rho(i) != 0)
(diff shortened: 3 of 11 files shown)


examples/exotica_examples/resources/configs/ik_solver_demo.xml
```diff
@@ -26,6 +26,8 @@
         </EndEffector>
       </EffFrame>
     </Maps>
+    <StartState>0 0 0 0 0 0 0</StartState>
+    <NominalState>0 0 0 0 0 0 0</NominalState>
```
wxmerkt: We may want to use the usual nominal pose here with the bent arm?

exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
```diff
@@ -56,13 +56,19 @@ class UnconstrainedEndPoseProblem : public PlanningProblem, public Instantiable<
     double getRho(const std::string& task_name);
     Eigen::VectorXd getNominalPose();
     void setNominalPose(Eigen::VectorXdRefConst qNominal_in);
+    virtual void preupdate();
```
wxmerkt: Perhaps we should add it to the documentation that preupdate needs to be called in the solver implementation and where.

exotica/src/Problems/UnconstrainedEndPoseProblem.cpp
```diff
@@ -84,16 +84,44 @@ void UnconstrainedEndPoseProblem::Instantiate(UnconstrainedEndPoseProblemInitial
...
+Eigen::VectorXd UnconstrainedEndPoseProblem::getScalarJacobian()
+{
+    return J.transpose()*S*ydiff*2.0;
 }
 
 void UnconstrainedEndPoseProblem::Update(Eigen::VectorXdRefConst x)
 {
     scene_->Update(x);
+    Phi.setZero(PhiN);
+    J.setZero();
```
wxmerkt: This is currently done in each individual taskmap. We should follow up and remove the unnecessary setting to zero there.

exotica/src/Problems/UnconstrainedTimeIndexedProblem.cpp
```diff
@@ -154,13 +170,25 @@ void UnconstrainedTimeIndexedProblem::Update(Eigen::VectorXdRefConst x, int t)
...
         if (Rho[t](i) != 0)
             Tasks[i]->update(x, Phi[t].data.segment(Tasks[i]->Start, Tasks[i]->Length), J[t].middleRows(Tasks[i]->StartJ, Tasks[i]->LengthJ));
     }
-    ydiff[t] = y[t] - Phi[t];
+    ydiff[t] = Phi[t] - y[t];
+}
+
+double UnconstrainedTimeIndexedProblem::getScalarCost(int t)
+{
+    return ydiff[t].transpose()*S[t]*ydiff[t];
```
wxmerkt: Make a check whether t is in the valid bounds and throw_pretty otherwise - see e.g. the getRho and setRho - I added it there because it's easy to accidentally be off by one and you don't want the Python kernel with all its state die because of that.

exotica/src/Problems/UnconstrainedTimeIndexedProblem.cpp
```diff
@@ -154,13 +170,25 @@ void UnconstrainedTimeIndexedProblem::Update(Eigen::VectorXdRefConst x, int t)
...
+}
+
+Eigen::VectorXd UnconstrainedTimeIndexedProblem::getScalarJacobian(int t)
+{
+    return J[t].transpose()*S[t]*ydiff[t]*2.0;
```
wxmerkt: Likewise, check whether t is valid.

exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
VladimirIvan: Is there a page in the docs where this can go?

examples/exotica_examples/resources/configs/ik_solver_demo.xml
VladimirIvan: I prefer zeros for simplicity. Any other value would require a comment explaining what pose that is.

exotica/include/exotica/Problems/UnconstrainedEndPoseProblem.h
wxmerkt: Not yet but you can create one

examples/exotica_examples/resources/configs/ik_solver_demo.xml
wxmerkt: Shouldn't nominal default to zeros or middle-of-joint-range if not set?

VladimirIvan: No, no nominal pose vector will result in not using the null space at all (min-norm).

PR #190, opened 2017-11-05
Fix of nearest contact points and work-around for primitive-vs-mesh collisions

### LINKED ISSUE #184: Contact points for collision between primitive shape and mesh wrong
This is the proxy for the closest distance between the red hip yaw link sphere and the torso yaw link mesh:

image

**Note:** This is the proxy being returned (distance seems correct):
```
Proxy: 'torsoYawLink_collision_0' - 'rightHipYawLink_collision_0', c1:          0 -0.0228953 -0.0911919 c1:          0 -0.0258317    0.92829 d: 0.00625062
```

wxmerkt:
- Splits the collision types into four pairs and fixes contact points for each of the four cases.
- As highlighted in #184, primitive-vs-mesh can't be fixed right now. Thus, a work-around flipping the order internally to mesh-vs-primitive and changing return values is implemented. 
- Validated with both LIBCCD and INDEP for primitive-sphere vs mesh-sphere and mesh-cube.

However, while this works fine for the simple use cases it doesn't yet work with the more complex meshes used for experiments.

Resolves #184

### DIFF
--- a/exotations/collision_scene_fcl_latest/include/collision_scene_fcl_latest/CollisionSceneFCLLatest.h
+++ b/exotations/collision_scene_fcl_latest/include/collision_scene_fcl_latest/CollisionSceneFCLLatest.h
@@ -80,7 +80,7 @@ public:
     /**
        * \brief Check if the whole robot is valid (collision only).
        * @param self Indicate if self collision check is required.
-       * @return True, if the state is collision free..
+       * @return True, if the state is collision free.
        */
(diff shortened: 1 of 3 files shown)


wxmerkt:
Just updated this PR with a new logic respecting all 16 different cases for collision pairs and swapping the narrowphase solver when meshes are involved (INDEP won't work reliably). I haven't swapped them for the normal binary collision check as I suspect both to work and we swapped it the opposite way there recently (can't recall or find notes on details/reasoning though).

PR #193, opened 2017-11-15
Indexed sampling problem update

yimingyanged:
add velocity limits to time indexed sampling problem

### DIFF
--- a/exotica/src/Problems/TimeIndexedSamplingProblem.cpp
+++ b/exotica/src/Problems/TimeIndexedSamplingProblem.cpp
@@ -57,10 +57,17 @@ void TimeIndexedSamplingProblem::Instantiate(TimeIndexedSamplingProblemInitializ
 {
     Parameters = init;
     goal_ = init.Goal;
-    T_ = init.T;
-    T_start_ = 0;
-    T_goal_ = init.T;
+    if (goal_.rows() != N)
+        throw_named("Dimension mismatch: problem N=" << N << ", but goal state has dimension " << goal_.rows());
+    T = init.T;
+    if (T < 0)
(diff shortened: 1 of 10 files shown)


exotica/src/Problems/TimeIndexedSamplingProblem.cpp
```diff
@@ -58,6 +58,9 @@ void TimeIndexedSamplingProblem::Instantiate(TimeIndexedSamplingProblemInitializ
     Parameters = init;
     goal_ = init.Goal;
     T_ = init.T;
+    T_start_ = 0;
+    T_goal_ = init.T;
+    vel_limits_ = init.JointVelocityLimits;
```
wxmerkt: Should check size - if 1 then apply same to all, if N OK, else throw_pretty

exotica/src/Problems/TimeIndexedSamplingProblem.cpp
```diff
@@ -102,13 +105,29 @@ void TimeIndexedSamplingProblem::Instantiate(TimeIndexedSamplingProblemInitializ
     }
 }
 
+Eigen::VectorXd TimeIndexedSamplingProblem::applyStartState()
+{
+    scene_->setModelState(startState, T_start_);
+    return scene_->getControlledState();
+}
+
```
VladimirIvan: Move T_start_ into the `PlanningProblem` and use the implementation above for `PlanningProblem::applyStartState()`

exotica/src/Problems/TimeIndexedSamplingProblem.cpp
VladimirIvan: Check vector size here.

Issue #186, opened 2017-11-03
Allow loaded scenes to be attached to a frame


wxmerkt:
Thus that they can be moved / transformed in the Scene as required

PR #199, opened 2017-11-16
Add support for loading scenes with custom transforms

wxmerkt:
A first step towards - and current solution/workaround for - #186


VladimirIvan:
Will this load offsets correctly if you re/load multiple scene files?

PR #207, opened 2017-12-05
Make UnconstrainedTimeIndexedProblem resizable during runtime

wxmerkt:
This PR will allow runtime resizing of the trajectory length without having to unnecessarily re-initialise the entire problem, e.g. when optimising over different time horizons.

- Makes ``T`` private and adds getters and setters
- Moves all users of ``T`` to the getter
- Make ``UnconstrainedTimeIndexedProblem`` resizable by re-initialising all required variables upon changes to the time horizon

### DIFF
--- a/exotations/solvers/aico/src/AICOsolver.cpp
+++ b/exotations/solvers/aico/src/AICOsolver.cpp
@@ -111,6 +111,9 @@ void AICOsolver::specifyProblem(PlanningProblem_ptr problem)
 
 void AICOsolver::Solve(Eigen::MatrixXd& solution)
 {
+    // Check if the trajectory length has changed, if so update the messages.
+    if (prob_->getT() != lastT) initMessages();
+
     prob_->resetCostEvolution(max_iterations);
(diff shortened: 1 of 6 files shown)


exotations/solvers/aico/src/AICOsolver.cpp
```diff
@@ -106,7 +106,6 @@ void AICOsolver::specifyProblem(PlanningProblem_ptr problem)
     MotionSolver::specifyProblem(problem);
     prob_ = std::static_pointer_cast<UnconstrainedTimeIndexedProblem>(problem);
 
-    T = prob_->T;
     initMessages();
```
VladimirIvan: ```initMessages()``` uses ```T``` to resize internal vectors. If you call ```Problem.setT()``` and then ```solve``` the solver will crash.

wxmerkt: Good catch - I've added c272d99 to address this.

PR #210, opened 2017-12-06
Fixes kinematic tree update

### LINKED ISSUE #206: Attached object does not get updated
When the environmental object has smaller ID than its parent, then it will not get updated.

VladimirIvan:
- Changes update of the kinematic tree to go in a topological order.
- Added missing tag into the time indexed config.

Fixes #206

### DIFF
--- a/examples/exotica_examples/resources/configs/time_indexed_sampling_demo.xml
+++ b/examples/exotica_examples/resources/configs/time_indexed_sampling_demo.xml
@@ -25,6 +25,7 @@
     <Maps>
       <CollisionCheck Name="Collision" SelfCollision="1" />
     </Maps>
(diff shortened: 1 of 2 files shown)


wxmerkt:
A few comments and clarifications:
- Does this support e.g. having the root link follow a trajectory? Because it's set to ``KDL::Frame()`` when the ID is 0. 
- Python tests failed on Travis
- Formatting not applied

VladimirIvan:
Points 1 and 2 are now fixed.
I thought formatting is applied automatically.

wxmerkt:
No, we didn't go for the automatic code formatting git hook in the end. We could add it including instructions to the documentation though and I am considering it. I simply run the documented command from the root of the tree just before committing (and have an alias in the bashrc).

PR #213, opened 2017-12-13
Expose validity checker resolution

VladimirIvan:
- Exposes time indexed rrt-connect validity check resolution via the initializer.
- Default value is ```<ValidityCheckResolution>0.01</ValidityCheckResolution>```

This should be propagated back to all OMPL solvers as well.

Merge #211 before merging this.


VladimirIvan:
0.01 is what OMPL uses by default. Just copied it over to keep the same settings.

PR #214, opened 2017-12-29
Fixes memory leaks in KinematicElement and Initializer python bindings

VladimirIvan:
- ```KinematicElement``` never gets deleted once its added to the tree because shared pointers to parents/children keep it alive. Changed `Parent` and `ClosestRobotLink` to a ```std::weak_ptr``` to fix this.
- Now, every time the `Parent` and `ClosestRobotLink` are used, they have to be converted to a shared pointer using ```Parent.lock()```.
- The ```Initializer``` class python wrapper constructs a tuple and a dictionary. Each entry into the tuple/dictionary was created anew but never freed (keeping these containers alive). The temporary python objects are now being freed (```Py_DECREF```) after being added to the container.


VladimirIvan:
Not after fixing yet. Still testing.

wxmerkt:
Would be great to see before/after of a test script to get an idea of improvement for future reference

PR #220, opened 2018-01-16
Refactor cost terms and add constrained problems

VladimirIvan:
- Added a ```Task``` class that contains a list of task maps. Each ```Task``` represents a set of cost terms, equality/inequality constraints, or validity check criteria. For each task, `rho` (precision/weight also stored in matrix form as `S`), `ydiff`, `y` (goal), `phi` (forward map), and `J` (Jacobian).
    - ```TimeIndexedTask``` contains a time indexed set of task maps+parameters.
    - ```EndPoseTask``` contains a set of task maps+parameters.
    - ```TimeIndexedTask``` contains a set of task maps+parameters except for `J`.
- Planning problems now use the corresponding `Task` type to store costs/constraints. E.g. ```UnconstrainedTimeIndexedProblem``` uses and updates the ```TimeIndexedTask Cost;```. An arbitrarily constrained end pose problem would then define: ```EndPoseTask Cost, Inequality, Equality;```. Examples have been updated for this (see the XML configs and the initialisers).
- Added ```BoundedEndPoseProblem```, ```BoundedTimeIndexedProblem```, ```EndPoseProblem```, ```TimeIndexedProblem``` - constrained versions of previously defined unconstrained problems.

TODO:
- Examples and tests using the new problems.
- Update the docs.


examples/exotica_examples/scripts/example_time_indexed_sampling.py
```diff
@@ -15,4 +15,6 @@
 
 Ts = solution[:,0]
 solution = solution[:,1:]
+print Ts
```
wxmerkt: This is Python 2 print

wxmerkt:
This actually breaks parallel build in an entirely fresh workspace. Single core build still appears to work but parallel (catkin default) is broken.

wxmerkt:
Just noticed: The BoundedEndPoseProblem doesn't have a nominal state

PR #228, opened 2018-01-23
Minor edits to the examples code

cmower:
Having gone through the _Basic_ tutorials I came across some minor issues in the code. 

* In Python examples I have changed
```
problem.setGoal(..,figureEight(), ..)
```
to 
```
goal=figureEight()
problem.setGoal(...,goal,...)
```
I think this is more readable and easier to explain in tutorials.

* The line `signal.signal(signal.SIGINT, sigIntHandler)` crops up in a few Python files causing the examples to throw an error. I have removed the offending lines from the respective examples however I suspect a more subtle problem caused this from `pybind`?

### DIFF
--- a/examples/exotica_examples/scripts/example_aico.py
+++ b/examples/exotica_examples/scripts/example_aico.py
@@ -17,7 +17,8 @@ for t in range(0,problem.T):
         problem.setRho('Frame',0.0,t)
     else:
         problem.setRho('Frame',1e5,t)
-        problem.setGoal('Frame',figureEight(t*problem.tau),t)
+        goal=figureEight(t*problem.tau)
+        problem.setGoal('Frame',goal,t)
 
 solution = solver.solve()
 
--- a/examples/exotica_examples/scripts/example_attach.py
+++ b/examples/exotica_examples/scripts/example_attach.py
@@ -40,7 +40,7 @@ for i in range(0,3):
 
 # Playback
 dt=0.03
-signal.signal(signal.SIGINT, sigIntHandler)
+
 while True:
     try:
         for i in range(0,len(solution)):
--- a/examples/exotica_examples/scripts/example_distance.py
+++ b/examples/exotica_examples/scripts/example_distance.py
@@ -11,7 +11,7 @@ sc.loadSceneFile('{exotica_examples}/resources/scenes/example_distance.scene')
 
 dt=0.01
 t=0.0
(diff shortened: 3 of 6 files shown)


examples/exotica_examples/scripts/example_aico.py
```diff
@@ -17,7 +17,8 @@ def figureEight(t):
         problem.setRho('Frame',0.0,t)
     else:
         problem.setRho('Frame',1e5,t)
-        problem.setGoal('Frame',figureEight(t*problem.tau),t)
+        goal=figureEight(t*problem.tau)
```
wxmerkt: This is not PEP8 compliant - it wasn't before but in edits we should respect that.

examples/exotica_examples/scripts/example_attach.py
```diff
@@ -40,7 +40,7 @@ def handleAttaching(i, x, scene):
 
 # Playback
 dt=0.03
-signal.signal(signal.SIGINT, sigIntHandler)
```
wxmerkt: This needs to stay - it's the only way to allow a graceful exit with Ctrl+C - otherwise you need to use ``kill` to kill Python.

examples/exotica_examples/scripts/example_distance.py
```diff
@@ -11,7 +11,7 @@
 
 dt=0.01
 t=0.0
-signal.signal(signal.SIGINT, sigIntHandler)
```
wxmerkt: Likewise, needs to stay. If it throws an error on your system please elaborate.

examples/exotica_examples/scripts/example_ik.py
```diff
@@ -20,10 +20,11 @@ def figureEight(t):
 t=0.0
 q=array([0.0]*7)
 print('Publishing IK')
-signal.signal(signal.SIGINT, sigIntHandler)
```
wxmerkt: Likewise, needs to stay. If it throws an error on your system please elaborate.

examples/exotica_examples/scripts/example_ik_noros.py
```diff
@@ -19,11 +19,12 @@ def figureEight(t):
 q=array([0.0]*7)
 print('Publishing IK')
 timer = exo.Timer()
-signal.signal(signal.SIGINT, sigIntHandler)
```
wxmerkt: Likewise, needs to stay. If it throws an error on your system please elaborate.

PR #230, opened 2018-01-23
Fix memory leak and exponentially slower updates

### LINKED ISSUE #225: KinematicElement memory leak / repeated updateSceneFrames causes problem updates to slow down dramatically
This is a major bug - if we frequently clean and load scenes the duration of a single problem update deteriorates very rapidly - for a repeated query of the exactly same IK request (226 problem updates) it initially takes 33ms and within a few seconds the same request (226 problem updates) takes >1s to solve.

Rough outline to reproduce:

- Clean Scene
- Load Scene From File
- Update problem goal ("y"), set model state, solve

I've isolated it to the interaction between scene cleaning and loading.

### Script to reproduce:
```python
import pyexotica as exo
import time
import numpy as np

ompl = exo.Setup.loadSolver(
    '{exotica_examples}/resources/configs/example_distance.xml')
sc = ompl.getProblem().getScene()

s = time.time()
ompl.getProblem().update(np.zeros(7,))
e = time.time()
print("Before:", e-s)

print("Clean and load scene 1000 times")
for _ in range(1000):
    sc.cleanScene()

s = time.time()
ompl.getProblem().update(np.ones(7,))
e = time.time()
print("After:", e-s)
```

### Output:
```
Before: 4.363059997558594e-05
Clean and load scene 1000 times
After: 0.00594019889831543
```

wxmerkt:
- Reduces incorrect use of ``std::shared_ptr`` and uses ``std::weak_ptr`` where possible (avoids cyclic references)
- Changes the main tree of the KinematicTree to ``ModelTree``, from ``Tree``
- Correctly clears references to links that are being deleted and removes the ``Children`` of a link that have been deleted - thereby decreasing the number of orphan elements updated and and thus fixing the increasing runtime
- Adds a destructor for ``KinematicElement`` such that it removes itself from its parent's ``Children`` list

NB: This is still not entirely thread-safe as technically one should check whether the ``weak_ptr.lock()`` actually returned a valid shared pointer.

Also, there is another memory leak in ``CollisionSceneFCLLatest`` only which gets triggered by collision queries (and slows things down a lot too) - I will track this down and address it later.

Test script run with both the current master and this PR (500 calls to ``cleanScene`` and ``loadSceneFile`` to trigger the ``updateSceneFrames``):

#### Current Master - CollisionSceneFCLLatest
```
Single Update Query Time Before: 3.695487976074219e-05  - Memory Usage Before: 65732
Single Update Query Time After: 0.0033359527587890625  - Memory Usage Before: 87940
>>> LEAK: 22208
```
#### Current Master - CollisionSceneFCL
````
Single Update Query Time Before: 2.5987625122070312e-05  - Memory Usage Before: 63516
Single Update Query Time After: 0.003376483917236328  - Memory Usage Before: 83228
>>> LEAK: 19712
````

#### This Fix - CollisionSceneFCLLatest
```
Single Update Query Time Before: 3.62396240234375e-05  - Memory Usage Before: 65840
Single Update Query Time After: 2.4318695068359375e-05  - Memory Usage Before: 65840
>>> LEAK: 0
```

#### This Fix - CollisionSceneFCL
```
Single Update Query Time Before: 2.3603439331054688e-05  - Memory Usage Before: 63636
Single Update Query Time After: 2.4557113647460938e-05  - Memory Usage Before: 63636
>>> LEAK: 0
```

Resolves #225

### DIFF
--- a/examples/exotica_examples/tests/issue225_kinematictree_memory_leak.py
+++ b/examples/exotica_examples/tests/issue225_kinematictree_memory_leak.py
@@ -27,6 +27,7 @@ memory_usage_intermediate = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
 print(">>> Intermediate:", " - Memory Usage Intermediate:",
       memory_usage_intermediate)
 
+print(">>> LEAK:", memory_usage_intermediate - memory_usage_before)
 assert (memory_usage_intermediate - memory_usage_before) == 0
 
 print(">>> updateSceneFrames 10000 times")
@@ -40,7 +41,7 @@ e = time.time()
 memory_usage_after = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
 print(">>> After:", e-s, " - Memory Usage After:", memory_usage_after)
 
+print(">>> LEAK:", memory_usage_after - memory_usage_before)
 assert (memory_usage_after - memory_usage_before) == 0
-print(">>> LEAK:", memory_usage_after-memory_usage_before)
 
 print('>>SUCCESS<<')
--- a/examples/exotica_examples/tests/issue225_kinematictree_memory_leak.py
+++ b/examples/exotica_examples/tests/issue225_kinematictree_memory_leak.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+import time
+import resource
+import numpy as np
+import pyexotica as exo
+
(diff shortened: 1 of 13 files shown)


wxmerkt:
While the leak is gone in Python 3, it is still present in Python @VladimirIvan :-/

PR #237, opened 2018-01-26
Fixes AICO's incorrect indexing into the Jacobian submatrices

VladimirIvan:
- Fixes AICO Jacobian indexing.
- Removes dependency of the `Task` classes on the `UnconstrainedTimeIndexedProblem`.
   - `TimeIndexedTask` now implements `reinitializeVariables(T, problem, phi)` method that sets up the time T.
   - The `phi` argument must be the same `TaskSpaceVector` as the one used to (and returned from the) initialize the task. Keep these as a member variable of the problem for problems with time indexing. E.g. `TimeIndexedProblem::CostPhi`. Problems that don't use the `TimeIndexedTask` can keep using a dummy local `TaskSpaceVector` (e.g. `EndPoseProblem::Cost`).
- AICO now prints out more details when in debug mode.

### DIFF
--- a/exotations/solvers/aico/src/AICOsolver.cpp
+++ b/exotations/solvers/aico/src/AICOsolver.cpp
@@ -133,6 +133,9 @@ void AICOsolver::Solve(Eigen::MatrixXd& solution)
         HIGHLIGHT("AICO::Solve called with initial trajectory guess");
     }
 
+    prob_->setStartState(q_init[0]);
+    prob_->applyStartState();
+
     Timer timer;
     if (debug_) ROS_WARN_STREAM("AICO: Setting up the solver");
     updateCount = 0;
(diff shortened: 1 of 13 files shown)


exotations/solvers/aico/src/AICOsolver.cpp
```diff
@@ -132,10 +133,6 @@ void AICOsolver::Solve(Eigen::MatrixXd& solution)
         HIGHLIGHT("AICO::Solve called with initial trajectory guess");
     }
 
-    prob_->preupdate();
-    prob_->setStartState(q_init[0]);
```
wxmerkt: This breaks warmstarts

wxmerkt:
Turns out this PR (rebase issue?) broke constraints and costs for reinitialisation in ``TimeIndexedProblem`` and ``BoundedTimeIndexedProblem`` (on every reinitialise, tasks got appended to the cost/constraints thus the size grew with duplicates).

Issue #242, opened 2018-01-30
AICO with initial trajectory: 0th/first step makes it worse


wxmerkt:
Note the cost for sweep -1 is 4538, the cost of sweep 0 is 49435. It gets worse from there.

```
[EXOTica]: AICO::Solve called with initial trajectory guess
[ WARN] [1517336856.485666794]: AICO: Setting up the solver
[ WARN] [1517336856.537988296]: Evaluating, sweep -1
[EXOTica]: Initial cost(ctrl/task/total): 2293.12/2245.11/4538.23, updates: 34
[ WARN] [1517336856.538078966]: AICO: Solving
[ WARN] [1517336856.589402968]: Evaluating, sweep 0
[EXOTica]: Sweep: 0, updates: 101, cost(ctrl/task/total): 632.658/48802.7/49435.4 (dq=0.67848, damping=0.01)
```

PR #245, opened 2018-01-31
AICO cleanup

VladimirIvan:
- Removed AICO dynamic mode.
- Removed unused variables in AICO (especially ones related to the dynamic mode).
- Removed AICO specific variables from the UnconstrainedEndPoseProblem.
- Updated examples.


wxmerkt:
Added a few commits to fix early exit (checking tolerance when first update hasn't been completed) and not undoing "bad" first sweep updates (cf. issue #242). Now works perfectly:

20180131_aico_fixed

Issue #223, opened 2018-01-18
Implement Hessians for task maps


wxmerkt:
Most of our existing taskmaps can be easily extended to provide Hessian information which we could and should exploit

PR #249, opened 2018-01-31
Adds Hessians to problems and Jdot to the kinematic tree

VladimirIvan:
- Adds the ```Hessian``` class (```Eigen::Array<Eigen::MatrixXd>```). Each element of the array is a NxN hessian for the respective task vector element.
- Problems own a Hessian for the task maps. Each Task then indexes into this in similarly to how Jacobians do.
- ```TaskMap``` computes an approximate Hessian (J'J) if the derived class doesn't provide one.
- Kinematic flags of a problem can be now changed in the initializer (-1=default, 0=FK only, 1=FK+Jacobian, 2=FK+Jacobian+Hessian/Jdot).
- Added planning problem tests that also test Jacobians and update Hessians where applicable.
- Fixed untested features in new problems.
- ```preupdate()``` call is no longer required in the solver. The problems handle this at init and in Rho updates.
- ```updateS``` in the tasks was using wrong indexing. Fixed.
- Relates to #223

TODO: 
- Implement actual Hessians for each task map.

### DIFF
--- a/exotica/src/TaskMap.cpp
+++ b/exotica/src/TaskMap.cpp
@@ -71,4 +71,15 @@ void TaskMap::taskSpaceDim(int& task_dim)
 {
     task_dim = taskSpaceDim();
 }
+
+void TaskMap::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::MatrixXdRef J, HessianRef H)
+{
+    update(x, phi, J);
+    H.resize(taskSpaceDim());
+    for(int i=0;i<taskSpaceDim();i++)
+    {
+        H(i)=J.row(i).transpose()*J.row(i);
+    }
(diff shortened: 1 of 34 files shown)


exotica/src/TaskMap.cpp
```diff
@@ -71,4 +71,14 @@ void TaskMap::taskSpaceDim(int& task_dim)
 {
     task_dim = taskSpaceDim();
 }
+
+void TaskMap::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::MatrixXdRef J, HessianRef H)
+{
+    update(x, phi, J);
+    H.resize(taskSpaceDim());
```
wxmerkt: Is this necessary on every update?

PR #253, opened 2018-02-01
Recreate kinematic solution when scene changes

VladimirIvan:
- Problem requests a kinematic solution for its task maps on initialisation. This solution maps to kinematic elements that become invalid when the scene is updated (removes elements). This PR fixes that.
- PlanningProblem requests frames and registers a callback to update the mapping for each TaskMap using ```PlanningProblem::updateTaskKinematics()```.
- Scene stores the request and a pointer to the callback. It call this callback upon request and then every time the scene objects get re-added.

Merge #249 before merging this.


exotica/src/PlanningProblem.cpp
```diff
@@ -179,6 +178,12 @@ void PlanningProblem::InstantiateBase(const Initializer& init_)
     }
 }
 
+void PlanningProblem::updateTaskKinematics(std::weak_ptr<KinematicResponse> response)
```
wxmerkt: This should be shared_ptr

exotica/src/Scene.cpp
```diff
@@ -550,7 +559,7 @@ void Scene::addTrajectory(const std::string& link, std::shared_ptr<Trajectory> t
     const auto& it = tree.find(link);
     if (it == tree.end()) throw_pretty("Can't find link '" << link << "'!");
     if (traj->getDuration() == 0.0) throw_pretty("The trajectory is empty!");
-    trajectory_generators_[link] = std::pair<std::shared_ptr<KinematicElement>, std::shared_ptr<Trajectory>>(it->second.lock(), traj);
+    trajectory_generators_[link] = std::pair<std::weak_ptr<KinematicElement>, std::shared_ptr<Trajectory>>(it->second, traj);
```
wxmerkt: This need to be fixed/updated in updateSceneFrames or else loading another scene can break trajectory-generated objects

VladimirIvan:
- Changed weak_ptr to shared_ptr in updateTaskKinematics.
- Fixed trajectory generators upon loading new scenes (through lazy updates).
- Made custom object and trajectory KinematicElement re-creation lazy. It now runs on updated if necessary. This means that a custom object/trajectory can use a non-existing frame as a parent as long as they get created before running the scene update. This means that the scene can be cleared and re-loaded from files without errors. Note that if a frame is actually missing, the exception will only get thrown when updating.
- Added a test for loading and re-loading the scene.

PR #258, opened 2018-02-03
Exposes attach object features in the XML

VladimirIvan:
- Added AttachLink initialiser.
- Scene uses these to attach objects to existing frames on init.
- EndPoseProblems now use ```tStart``` when updating the scene. This means that an IK solver can now make use of trajectories.
- Added example using all of the above.


exotica/src/Scene.cpp
```diff
@@ -122,6 +126,19 @@ void Scene::Instantiate(SceneInitializer& init)
     updateSceneFrames();
     updateInternalFrames(false);
 
+    for (const exotica::Initializer& linkInit : init.AttachLinks)
```
edbot: Does this need to be re-attached when scenes are re-loaded or cleaned / updateSceneFrames is called?

VladimirIvan: Absolutely. The commit below fixes that. Thanks for spotting it.

PR #286, opened 2018-04-18
Newer Eigen version for pybind11

christian-rauch:
This PR adds the package [`eigen_catkin`](https://github.com/ethz-asl/eigen_catkin) (and its dependency [`catkin_simple`](https://github.com/catkin/catkin_simple)) to the repo and as dependency to catkin packages that generate pybind11 bindings. With `eigen_catkin` there is no need to install newer Eigen versions system-wide (thus avoiding nasty side effects). Tested on Ubuntu 14.04.

Using newer compiler versions from https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test and compiling via: `catkin build -DCMAKE_CXX_COMPILER=/usr/bin/g++-7` also allows to use a newer compiler just for this workspace.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,9 @@
 [submodule "fcl_catkin"]
 	path = fcl_catkin
 	url = https://github.com/ipab-slmc/fcl_catkin.git
+[submodule "eigen_catkin"]
+	path = eigen_catkin
+	url = https://github.com/ethz-asl/eigen_catkin.git
+[submodule "catkin_simple"]
+	path = catkin_simple
+	url = https://github.com/catkin/catkin_simple.git
--- a/exotations/solvers/ompl_solver/CMakeLists.txt
+++ b/exotations/solvers/ompl_solver/CMakeLists.txt
@@ -6,6 +6,7 @@ find_package(OMPL REQUIRED)
 find_package(catkin REQUIRED COMPONENTS
   exotica
   exotica_python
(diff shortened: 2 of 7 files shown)


wxmerkt:
Thank you very much for the contribution. In principal, I'd agree this is a better way, however, I'd prefer if these could be installed from binary (they aren't released from upstream it seems) rather than adding them as submodules - this can lead to double-packages and clashes if used in other packages in the workspace. I've opened issues upstream but there are reservations from the ROS side w.r.t. releasing ``catkin_simple`` or using it on the buildfarm. Perhaps, a ``catkin_simple``-free ``eigen_catkin``-wrapper would be the solution. As it stands, adding this dependency will complicate releasing exotica as binaries down the road.

Further, and unrelated, we need to fix some of our Eigen-related code (aligned_allocator in STL containers, only passing references).

VladimirIvan:
I agree with Wolf. While it would be good to have proper package dependency on a specific version of eigen, the `catkin_eigen ` solution is not very standard.

Also, making 14.04 more compatible is not very high on the priority list. It is at the end of its lifetime. The system version of eigen on 16.04 works so this PR just adds submodules there.

christian-rauch:
I agree that the added submodule would only solve a dependency issue for an old Ubuntu LTS. I will close the PR but keep the patches as reference for other users of older LTS versions.

PR #289, opened 2018-04-24
Get objective function value per solver iteration

christian-rauch:
This PR enables to retrieve the costs of the objective function for each iteration and the final solution.
- move `double error` to solver base class (assuming it will be used by inheriting solvers)
- implement cost evolution for IK solver

### DIFF
--- a/exotations/solvers/ik_solver/src/ik_solver/IKSolver.cpp
+++ b/exotations/solvers/ik_solver/src/ik_solver/IKSolver.cpp
@@ -126,6 +126,8 @@ int IKsolver::getLastIteration()
 
 void IKsolver::Solve(Eigen::MatrixXd& solution)
 {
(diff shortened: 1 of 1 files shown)


wxmerkt:
Thank you for your contribution and looking into this. I believe what you are looking for is ``ydiff`` of the corresponding cost task - which, for the case of the ``UnconstrainedEndPoseProblem`` may not be exposed to Python anymore at the moment. I.e. the ``ydiff`` of the ``Cost`` should correspond to the error. The Cost itself is ``ydiff.T*S*ydiff``. Happy to stop by and discuss later :-)

christian-rauch:
I reversed the movement of `double error` to the base class.
As discussed, the squared weighted cost is replaced by a call to `getScalarCost`.

PR #292, opened 2018-04-30
Point2Line task map

christian-rauch:
This PR:
- fixes an issue where comments in *.in files including a semicolon would be recognised as code
- inhibits `Setting maximum iterations to...` when the debug flag is not set
- adds a `Point2Line` task map

The `Point2Line` task map computes the vector from a point in the Link frame to its closest point on a line in the Base frame. The magnitude of this vector is used for the task space error phi, and its direction is used to update the Jacobian of the objective.


christian-rauch:
Thanks, I missed the formatting page in the documentation completely. I merged the changes into the last commit.

PR #306, opened 2018-05-09
Move MaxIterations to MotionSolver

wxmerkt:
This takes #291 one step further and actually moves the MaxIterations property into MotionSolver - and thus reduces the need for repeated setting of the maximum iterations variable in every solver. It also thereby reduces the risk of missing to implement this functionality in some solvers while it's present in others.


wxmerkt:
I already exposed it three months ago in commit https://github.com/ipab-slmc/exotica/commit/37f8bd58fe9ff38cdaee19a52af8d90a095656e0

Cf. https://ipab-slmc.github.io/exotica/Python-API.html#pyexotica.MotionSolver

PR #309, opened 2018-05-24
Scene add removeObject and fix segfault after loading scenes/adding objects

wxmerkt:


VladimirIvan:
This changes the behaviour from lazy internal frames updates to updating after every call load/addObject/...
Is that correct?

This will a) reinitialize internal frames every time a new object is added, b) the initializer scene object loading order depends on this.
Why is the change necessary?

wxmerkt:
Without it loading multiple scenes without manually calling the updateInternalFrames result in a segfault. Also, without a scene or problem update we would otherwise publish incorrect information. It needs to be updated after updating new objects anyways and I tested against the loading order (i.e. check the test passing).

VladimirIvan:
```updateInternalFrames``` can now be called from python without calling ```updateSceneFrames```. This causes a segfault. Perhaps remove ```updateInternalFrames``` from the python wrapper?
Otherwise good to go.

PR #325, opened 2018-06-18
Change debug messages in IKsolver and MotionSolver

christian-rauch:
- add debug message for IKsolver
- remove debug message for MotionSolver

I remove the message for MotionSolver, since its a configurable parameter that is changed by the user. The message is therefore not informative.


wxmerkt:
Looks good and I will merge - I kept the message about the setting maximum iterations before as some solvers or initialisers set/adjusted it internally without the user being aware.

PR #335, opened 2018-06-24
Added quasistatic task map

VladimirIvan:
- New task map implementing signed distance of CoM inside a support polygon.
- The CoM is computed using all robot links.
- The support polygon is computed using quickhull from specified frames projected on the XY plane.
- The signed distance is a harmonic potential field based on the electrostatic field. The values inside the support polygon are negative and outside are positive. Inside/outside is distinguished using the winding number.
- The `PositiveOnly` caps the values to zero inside the support polygon. Use this for unconstrained optimisation. Inequality constraints can use the full equation.
- The debug mode will enable publishing a marker array showing the CoM and the support polygon.
- Test and example have been added.
- The example script requires the `val_description` package to be compiled but this is not an explicit dependency of the repo.

### DIFF
--- a/exotations/task_maps/task_map/include/task_map/ConvexHull.h
+++ b/exotations/task_maps/task_map/include/task_map/ConvexHull.h
@@ -1,0 +1,85 @@
+#ifndef CONVEXHULL_H
+#define CONVEXHULL_H
+
+#include "exotica/Tools.h"
+#include "Eigen/Dense"
+
+namespace exotica
+{
+
+double lineDist2D(Eigen::VectorXdRefConst p1, Eigen::VectorXdRefConst p2, Eigen::VectorXdRefConst p)
+{
+    return (p(1) - p1(1)) * (p2(0) - p1(0)) -
+           (p2(1) - p1(1)) * (p(0) - p1(0));
+}
+
+
+std::list<int> quickHull(Eigen::MatrixXdRefConst points, std::list<int>& halfPoints, int p1, int p2)
+{
+    int ind = -1;
+    double max_dist = 0;
+    std::list<int> newHalfPoints;
+    for (int i : halfPoints)
+    {
+        double d = lineDist2D(points.row(p1), points.row(p2), points.row(i));
+        if (d>0.0)
+        {
+            newHalfPoints.push_back(i);
+        }
+        if (d > max_dist)
+        {
+            ind = i;
+            max_dist = d;
+        }
+    }
+
+    std::list<int> hull;
+
+    if (ind == -1)
+    {
+        hull.push_back(p2);
+    }
+    else
+    {
+        hull.splice(hull.begin(), quickHull(points, newHalfPoints, p1, ind));
+        hull.splice(hull.end(), quickHull(points, newHalfPoints, ind, p2));
+    }
+    return hull;
+}
+
+std::list<int> convexHull2D(Eigen::MatrixXdRefConst points)
+{
+    if (points.cols()!=2) throw_pretty("Input must contain 2D points!");
+
+    int n = points.rows();
+
+    std::list<int> hull;
+    std::list<int> halfPoints;
+
+    if (n < 3)
+    {
+        for(int i=0; i<n; i++) hull.push_back(i);
+    }
+    else
+    {
+        int min_x = 0, max_x = 0;
+        halfPoints.push_back(0);
+        for (int i=1; i<n; i++)
+        {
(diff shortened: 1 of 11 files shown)


wxmerkt:
@VladimirIvan The Jacobian tests fail on 14.04/Indigo with error out of bounds, e.g.

```
[EXOTica]: QuasiStatic test inside
Testing random configurations:
Test passed
Testing Jacobian:
[EXOTica]: x:  0.96307 0.206424 0.983471
[EXOTica]: J*:
  -0.90043  -0.214399 -0.0505608
[EXOTica]: J:
 -0.900446  -0.214401 -0.0505615

Test test_maps failed
terminate called after throwing an instance of 'exotica::Exception'
  what():  In /home/wxm/dev/exotica_ws/src/exotica/examples/exotica_examples/tests/test_maps.cpp
testJacobian 122
Jacobian error out of bounds: 1.55887e-05
```

Relaxing the ``eps`` does not resolve the problem.

exotations/task_maps/task_map/include/task_map/ConvexHull.h
```diff
@@ -0,0 +1,80 @@
...
+#include "exotica/Tools.h"
+
+namespace exotica
+{
+double lineDist2D(Eigen::VectorXdRefConst p1, Eigen::VectorXdRefConst p2, Eigen::VectorXdRefConst p)
+{
+    return (p(1) - p1(1)) * (p2(0) - p1(0)) -
+           (p2(1) - p1(1)) * (p(0) - p1(0));
+}
+
```
ferrolho: Hmm, is this function working as it is supposed to, @VladimirIvan ?

```python
>>> def lineDist2D(p1, p2, p):
...     return (p[1] - p1[1]) * (p2[0] - p1[0]) - (p2[1] - p1[1]) * (p[0] - p1[0])
... 
>>> lineDist2D([0, 0], [0, 0], [1, 1])
0
>>> lineDist2D([0, 0], [2, 0], [1, 1])
2
>>> lineDist2D([0, 0], [-2, 0], [1, 1])
-2
>>> lineDist2D([-2, 0], [0, 0], [1, 1])
2
>>> lineDist2D([0, 0], [1, 0], [1, 1])
1
```

VladimirIvan: Oh boy, you're right. It's supposed to be:
```
def lineDist2D(p1, p2, p):
  return ((p[1] - p1[1]) * (p2[0] - p1[0]) - (p2[1] - p1[1]) * (p[0] - p1[0]))/math.sqrt( (p2[0]-p1[0])**2 + (p2[1]-p1[1])**2 )
```

Classic vlad

ferrolho:
```python
def lineDist2D(p1, p2, p):
    num = (p[1] - p1[1]) * (p2[0] - p1[0]) - (p2[1] - p1[1]) * (p[0] - p1[0])
    denom = math.sqrt((p2[0] - p1[0]) ** 2 + (p2[1] - p1[1]) ** 2)
    return num / denom
```

This does not handle some corner cases. For example:
```python
>>> lineDist2D([0, 0], [0, 0], [1, 1])Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 4, in lineDist2D
ZeroDivisionError: float division by zero
```

Also, the lack of documentation or a more descriptive name makes me a bit confused. Is the function supposed to compute the shortest distance from a given point to any point in the line? Or does it compute the *y*-distance? Here's why I ask:
```python
>>> lineDist2D([0, 0], [1, 0], [2, 0])
0.0
```

VladimirIvan: What I was aiming for was perpendicular oriented distance be tween a point and a line. It is actually not defined when the line points are coincident. The division by zero can be guarded and return 0.

Perhaps you got it confused with a distance to a line segment. That can't be signed and it is not what's needed for the quickhull algorithm.

Although, I have just realized that you actually don't need to divide by `||p2-p1||` for the quickhull to work. It does need a new name though. I believe it's called a [determinant in that case](http://mathworld.wolfram.com/CrossProduct.html).

ferrolho: Something like this, right?
img

wxmerkt:
Do the contact points need to be entered in order?

image

PR #349, opened 2018-07-04
Upgrade FCL to include EPA bug fixes

wxmerkt:
Brings in the following changes to FCL: https://github.com/flexible-collision-library/fcl/compare/cd7a0c1fcb461b522f7f1a397c86fa964639462c...07c8480f70e4b5bfa15cb37b76fc5271d2502e82

- Fixes the simplex extraction (no more segfaults, hopefully)
- Fix EPA


wxmerkt:
Superseded by #365 which includes many more bug fixes and corrections to contact/nearest points, distances, and simplifies the logic significantly while fixing long-standing issues with transforms not being updated (ouch!).

PR #351, opened 2018-07-08
Fix segmentation fault when using joint limits

wxmerkt:
Since 86894b343c898d4f8b5e397b404c9f0cb9df0fec using the ``JointLimit`` task map or directly indexing the column of ``KinematicTree::getJointLimits().col(0)`` caused an assignment segmentation fault. While I do not fully understand why changing the type to a copy (returning a ``Eigen::MatrixXd``) does not resolve the issue, the changes here fix the wrongly introduced ``Eigen::MatrixXdRef`` return type that should only be used as a function parameter (my bad).

In order to reproduce the before case, you can simply add a ``JointLimit`` taskmap to the IK example. It worked before 86894b343c898d4f8b5e397b404c9f0cb9df0fec and was broken thereafter.

### DIFF
--- a/exotations/task_maps/task_map/src/JointLimit.cpp
+++ b/exotations/task_maps/task_map/src/JointLimit.cpp
@@ -59,14 +59,18 @@ void JointLimit::Initialize()
 {
     double percent = init_.SafePercentage;
 
-    std::vector<std::string> jnts = scene_->getJointNames();
-    N = jnts.size();
+    N = scene_->getSolver().getNumControlledJoints();
+
+    // TODO: Strictly speaking this is incorrect as joint limits can be changed during runtime.
     Eigen::MatrixXd limits = scene_->getSolver().getJointLimits();
-    low_limits_ = limits.col(0);
-    high_limits_ = limits.col(1);
+
+    low_limits_.resize(N);
+    high_limits_.resize(N);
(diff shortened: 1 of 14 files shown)


exotations/task_maps/task_map/src/JointLimit.cpp
```diff
@@ -59,14 +59,18 @@ void JointLimit::Initialize()
 {
     double percent = init_.SafePercentage;
 
-    std::vector<std::string> jnts = scene_->getJointNames();
-    N = jnts.size();
+    N = scene_->getSolver().getNumControlledJoints();
+
+    // TODO: Strictly speaking this is incorrect as joint limits can be changed during runtime.
     Eigen::MatrixXd limits = scene_->getSolver().getJointLimits();
-    low_limits_ = limits.col(0);
```
wxmerkt: I am unclear why this now would result in a runtime segmentation fault independent of whether ``getJointLimits()`` returns an ``Eigen::MatrixXd`` (copy), a ``const Eigen::MatrixXd&``, or a ``Eigen::MatrixXd&``.

VladimirIvan: Let's keep an eye on these Eigen issues.
It looks good otherwise.

PR #352, opened 2018-07-09
CollisionSceneFCL: suppress version output at every contruction

christian-rauch:
This simply removes the output `FCL version:` at every construction of `CollisionSceneFCL` or `CollisionSceneFCLLatest`. It does not change during runtime anyway.

### DIFF
--- a/exotations/collision_scene_fcl/include/collision_scene_fcl/CollisionSceneFCL.h
+++ b/exotations/collision_scene_fcl/include/collision_scene_fcl/CollisionSceneFCL.h
@@ -67,6 +67,8 @@ public:
        */
     virtual ~CollisionSceneFCL();
 
+    void setup();
+
     static bool isAllowedToCollide(fcl::CollisionObject* o1, fcl::CollisionObject* o2, bool self, CollisionSceneFCL* scene);
     static bool collisionCallback(fcl::CollisionObject* o1, fcl::CollisionObject* o2, void* data);
     static bool collisionCallbackDistance(fcl::CollisionObject* o1, fcl::CollisionObject* o2, void* data, double& dist);
--- a/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
+++ b/exotations/collision_scene_fcl/src/CollisionSceneFCL.cpp
@@ -47,7 +47,7 @@ namespace exotica
 {
 CollisionSceneFCL::CollisionSceneFCL()
 {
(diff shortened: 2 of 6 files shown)


wxmerkt:
Can we wrap this in ``if (scene_->debug_)`` instead of commenting out?

christian-rauch:
I can neither find a variable `scene_` nor `debug_`.

wxmerkt:
Ah, I see that's correct. Perhaps we should pass the Scene's debug flag on instantiation to the CollisionScene then.

christian-rauch:
I am forwarding the debug flag to `CollisionScene`. But when setting the flag via
```XML
    <PlanningScene>
      <Scene>
        <Debug>1</Debug>
      </Scene>
    </PlanningScene>
```
I am getting:
```
[EXOTica]: [] Running in debug mode, planning scene will be published to 'EXOTicaServer//PlanningScene'
[EXOTica]: [] Exotica Scene initialized
```
but it's not forwarded to `CollisionSceneFCL`.

Any ideas why?

wxmerkt:
Try the debug on "PlanningScene" instead of "Scene".

christian-rauch:
This:
```XML
    <PlanningScene>
      <Debug>1</Debug>
      <Scene>
        <JointGroup>world_arm_hand</JointGroup>
        <URDF>{lwr_defs}/robots/lwr_sdh.urdf</URDF>
        <SRDF>{kuka_lwr_sdh2_moveit}/config/lwr.srdf</SRDF>
        <!--<Debug>0</Debug>-->
      </Scene>
    </PlanningScene>
```

gives me an error about an undefined joint group:
```
[ERROR] [1531235039.800387204]: Robot semantic description not found. Did you forget to define or remap '/robot_description_semantic'?
[ERROR] [1531235039.837980605]: Group '' not found in model 'lwr'
Joint group '' not defined in the robot model!
```

VladimirIvan:
The `<Debug>1</Debug>` tag has to be within `<Scene></Scene>` as you had it originally.

It didn't print because the constructor gets called before you set debug flag to true.
Maybe move the print into a separate virtual public method and call it from the scene if the scene is in debug mode?

christian-rauch:
Yes off course, in this case the initialisation happens after the construction :-)
I added a setup method for additional constructions.

christian-rauch:
I see, but this would limit the functionality to debugging cases.

You need to add a new method to `CollisionScene` anyway and the `setup` method allows to be more flexible about post-construction cases. And it allows you to use `CollisionScene`'s debug flag in other places.

PR #355, opened 2018-07-09
OMPL console bridge fix

VladimirIvan:
The OMPL packages fail compilation because the console bridge API changed in the newer version.
This fix adds backward compatibility.
Hopefully this will get deprecated soon.


wxmerkt:
Which version of OMPL do you use on what system? I haven't seen this with stock-kinetic and @ferrolho managed to compile on melodic.

wxmerkt:
Alright, I see. As our CI is still on Indigo I think it would be good to add a unit test running OMPL with a fixed seed and comparing the plan. On the other hand, we fully compile Indigo on CI so perhaps see if ``apt-get update`` has a newer version?

VladimirIvan:
Turns out this is already deprecated. My test machine just needed dist upgrade.

PR #357, opened 2018-07-10
Replace pybind11_catkin with submodule

wxmerkt:


wxmerkt:
Caveat: users who update may need to delete the pybind11_catkin folder and then run ``git submodule update --init --recursive``.

PR #361, opened 2018-08-13
Alpha vector for IKsolver

christian-rauch:
This PR changes the scalar alpha to a vector to apply scales elementwise per state variable instead of scaling the whole state vector. The vector size must be 1 or N (state size).

### DIFF
--- a/exotations/solvers/ik_solver/init/IKsolver.in
+++ b/exotations/solvers/ik_solver/init/IKsolver.in
@@ -3,4 +3,4 @@ Optional double Tolerance = 1e-5;
 Optional double Convergence = 0.0;
 Optional double MaxStep = 0.02;
 Optional double C = 0.0;
-Optional double Alpha = 1.0;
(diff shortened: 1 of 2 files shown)


VladimirIvan:
What do you use this for?
There is Rho, which allows you to scale each task.
And there is W which allows you to scale the joint space metric (not exactly in the same way as the alpha scaling you propose though).

Alpha vector like that can add skew to the gradient and create suboptimal results and divergence.
Is this really what you need?

christian-rauch:
I am using the alpha vector in an on/off fashion to inhibit the optimisation of the 6D pose. Setting W to high values for these states still gives me results where the start pose diverges from the estimated pose. Would it make a difference if I apply alpha on the inverse Jacobian?

VladimirIvan:
I think that should be OK then. Just be aware that alpha is a convergence rate. This is misappropriating it for a problem specific weight. If I were to be pedantic, it is a hack but I shall allow it this time.

PR #363, opened 2018-08-16
Eigen::Vector templating and error checking

christian-rauch:
This PR changes `parseVector` into a template to allow different typed and size `Eigen::Vector`s. This reduces the amount of boilerplate code for different kind of `Eigen::Vector<SIZE><TYPE>`.

Additionally, it replaces the number conversion from streaming (`>>`) to the use of functions that throw exceptions when invalid numbers occur. This has the advantage that invalid numbers can be represented by `NAN` instead of returning a shortened vector, as done currently. However, the template `to_number` needs to be specialised for the remaining numerical types.


exotica/include/exotica/Tools/Conversions.h
```diff
@@ -161,40 +161,50 @@ inline std::string trim(const std::string& s)
...
-        ret(i++) = temp_entry;
-        text_parser >> temp_entry;
+        ret.conservativeResize(++i);
+        try
+        {
+            ret[i - 1] = to_number<T>(temp_entry);
+        }
+        catch (std::invalid_argument)
+        {
+            ret[i - 1] = NAN;
```
wxmerkt: ``NAN`` evaluates to ``quiet_NaN`` so perhaps best to use that one, cf. [here](https://en.cppreference.com/w/cpp/types/numeric_limits/quiet_NaN).

Issue #364, opened 2018-08-17
List of broken collision distance pairs


wxmerkt:
As of August 2018, the following pairs of distance queries are not expected to work with ``CollisionSceneFCLLatest``:

- [ ] Mesh vs Sphere - at distance [not implemented in FCL]
- [ ] Mesh vs Sphere - in collision [not implemented in FCL]
- [ ] Mesh vs Box - in collision [not implemented in FCL]
- [ ] Mesh vs Cylinder - in collision [not implemented in FCL]
- [ ] Mesh vs Mesh - in collision [implemented, but results wrong]

PR #365, opened 2018-08-17
Updates to CollisionScenes for distance computations, bug fixes, etc.

wxmerkt:
A lot of updates and fixes to the collision scene and especially distance and contact point computation since ~May. This is broken out of a larger collection of updates and only relates to CollisionScene-related changes.

- Includes an update of FCL to 0.5.94 which fixes many bugs, including in the GJK and EPA algorithms and that most (not all - primitive vs mesh being the notable exception) points are returned in world frame, and includes some new custom solver implementations for primitives.
- Changes to using exact points rather than shape centres where possible, and makes use of collision checks for distances instead of distance checks when in collision (much faster!).
- Documents which tests are broken (cf. #364) and fixes some of our internal bugs. 
- Simplifies distance checking logic due to upstream bug fixes.
- Fixes compilation of exotica in install-configured workspaces with external dependencies.

The correlated changes to FCL which this PR brings in are these: https://github.com/flexible-collision-library/fcl/compare/cd7a0c1fcb461b522f7f1a397c86fa964639462c...f15ffc21c9c0f6f6ab7d5369e8889ccbc59c2233


examples/exotica_examples/tests/collision_scene_distances.py
```diff
@@ -66,16 +67,17 @@ def testSphereVsSpherePenetrating(collisionScene):
 
     p = scene.getCollisionDistance("A", "B")
     debugPublish(p, scene)
+    print(p)
```
VladimirIvan: Is this required?
Also in subsequent pints?

examples/exotica_examples/tests/collision_scene_distances.py
```diff
@@ -247,8 +251,8 @@ def testSphereVsMeshPenetrating(collisionScene):
     print('PrimitiveSphere_vs_Mesh_Penetrating: isStateValid(True): PASSED')
 
     p = scene.getCollisionDistance("A", "B")
-    print(p)
     debugPublish(p, scene)
+    print(p) # qqq
```
VladimirIvan: Remove comment

examples/exotica_examples/tests/collision_scene_distances.py
wxmerkt: Yes. I left it in the ones that frequently break to be able to spot more easily.

Issue #372, opened 2018-09-01
python transformation get functions inconsistent


cmower:
* `getTranslation` returns a 3-vector as expected
* `getQuaternion` returns a 7-vector that is the translation and quaternion
* `getRPY` returns a 6-vector that is the translation and rpy angles
```
In [5]: (sol, prob)=exo.Initializers.loadXMLFull(XML)

In [6]: problem=exo.Setup.createProblem(prob)
[ INFO] [1535836557.597380920]: Loading robot model 'lwr'...
[EXOTica]: No maps were defined!

In [7]: scene = problem.getScene()

In [8]: scene.Update([0,0,0,0,0,0,0])

In [9]: t = scene.fk('lwr_arm_7_link')

In [10]: t
Out[10]: KDL::Frame ([0.000000 0.000000 1.198500] [0.000000 0.000000 1.000000 -0.000000])

In [11]: t.getTranslation()
Out[11]: array([ 0.    ,  0.    ,  1.1985])

In [12]: t.getQuaternion()
Out[12]: 
array([  0.00000000e+00,   0.00000000e+00,   1.19850000e+00,
         0.00000000e+00,   0.00000000e+00,   1.00000000e+00,
        -1.03411554e-13])

In [13]: len(t.getQuaternion())
Out[13]: 7

In [14]: t.getRPY()
Out[14]: 
array([ 0.        ,  0.        ,  1.1985    ,  0.        , -0.        ,
       -3.14159265])

In [15]: t.getTranslation()
Out[15]: array([ 0.    ,  0.    ,  1.1985])

```

cmower:
* It doesn't make sense that `getTranslation` returns only the translation when `getQuaternion`/`getRPY` return translation plus quaternion/rpy 
* `getQuaternion` should return `[x, y, z, w]` or `[w, x, y, z]` (what ever KDL has as its convention)
* `getRPY` should return `[r, p, y]`

PR #373, opened 2018-09-02
Fixing inconsistent get functions

cmower:
**Problem**: 
As described in #372 the get functions, listed below, used to return a translation + rotation where the rotation is in the given format. This is doesn't make sense; i.e. why return the translation when you request only the quaternion? This is also inconsistent with KDL, see [here](http://docs.ros.org/hydro/api/orocos_kdl/html/classKDL_1_1Rotation.html#a37f92d1b2c5e9708771543c521c276c4).

**Changes**:
* getQuaternion, now returns a 4-vector
* getRPY, now returns a 3-vector
* getZYZ, "
* getZYX, "
* getAngleAxis, "

Changes build successfully.

**Example (before)**:
```
ln [5]: scene.Update([0,0,0,0,0,0,0])

In [6]: t = scene.fk('lwr_arm_7_link')

In [7]: t.getTranslation()
Out[7]: array([ 0.    ,  0.    ,  1.1985])

In [8]: t.getQuaternion()
Out[8]: 
array([   0.    ,  0.    ,  1.1985, 0.00000000e+00,   0.00000000e+00,   1.00000000e+00,
        -1.03411554e-13])

In [9]: t.getRPY()
Out[9]: array([  0.    ,  0.    ,  1.1985, 0.        , -0.        , -3.14159265])

In [10]: t.getZYZ()
Out[10]: array([ 0.    ,  0.    ,  1.1985, -3.14159265,  0.        ,  0.        ])

In [11]: t.getZYX()
Out[11]: array([ 0.    ,  0.    ,  1.1985, -3.14159265, -0.        ,  0.        ])

In [12]: t.getAngleAxis()
Out[12]: array([  0.    ,  0.    ,  1.1985, 0.        ,  0.        ,  3.14159265])
```
**Example (after)**:
```
ln [5]: scene.Update([0,0,0,0,0,0,0])

In [6]: t = scene.fk('lwr_arm_7_link')

In [7]: t.getTranslation()
Out[7]: array([ 0.    ,  0.    ,  1.1985])

In [8]: t.getQuaternion()
Out[8]: 
array([  0.00000000e+00,   0.00000000e+00,   1.00000000e+00,
        -1.03411554e-13])

In [9]: t.getRPY()
Out[9]: array([ 0.        , -0.        , -3.14159265])

In [10]: t.getZYZ()
Out[10]: array([-3.14159265,  0.        ,  0.        ])

In [11]: t.getZYX()
Out[11]: array([-3.14159265, -0.        ,  0.        ])

In [12]: t.getAngleAxis()
Out[12]: array([ 0.        ,  0.        ,  3.14159265])
```


exotica_python/src/Exotica.cpp
```diff
@@ -574,7 +574,7 @@ PYBIND11_MODULE(_pyexotica, module)
     kdlFrame.def("getAngleAxis", [](KDL::Frame* me) { return getRotationAsVector(*me, RotationType::ANGLE_AXIS); });
     kdlFrame.def("getQuaternion", [](KDL::Frame* me) { return getRotationAsVector(*me, RotationType::QUATERNION); });
     kdlFrame.def("getTranslation", [](KDL::Frame* me) { Eigen::Vector3d tmp; for (int i = 0; i < 3; i++) { tmp[i] = me->p.data[i]; } return tmp; });
-    kdlFrame.def("get", [](KDL::Frame* me) { return getFrame(*me); });
+    kdlFrame.def("getFrame", [](KDL::Frame* me) { return getFrame(*me); });
```
wxmerkt: Also, please ask Vlad about this change I am not sure it does what's intended in the first place.

cmower:
Point 1: Spoke to Vald, he agrees with format getTranslationAnd[Rotation], will look for code in other places of exotica where original format is being called and update 
Point 2: Will do, thanks for the tutorial.

exotica_python/src/Exotica.cpp
cmower: Have asked and he agrees on getFrame, will update elsewhere where this is being called (I'm assuming in exotica_example?), will do some searching.

wxmerkt: It is not being used as far as I can tell and that is what was my point - what is this function meant to do? You can already instantiate from another frame, you can also get all data with the other methods, and you can instantiate from Eigen::Vectors or rotation matrices? @VladimirIvan

cmower:
* Formatting, as in exotica documentation, completed. 
* Build succeeds and outputs are as expected when tested in python.
* All commits squashed into two logical change-sets.

PR #376, opened 2018-09-26
fix non-catkin ompl dependency

christian-rauch:
I get the error:
```
CMake Error at [...]/exotica/exotations/solvers/time_indexed_rrt_connect/CMakeLists.txt:4 (find_package):
  By not providing "FindOMPL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OMPL", but
  CMake did not find one.
```
Renaming the dependency correctly to `ompl` (lower case) and setting it as system dependency solve this issue.

Note: This is on Ubuntu 18.04 / ROS melodic

### DIFF
--- a/exotations/solvers/ompl_solver/CMakeLists.txt
+++ b/exotations/solvers/ompl_solver/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ompl_solver)
 
-find_package(OMPL REQUIRED)
+find_package(ompl REQUIRED)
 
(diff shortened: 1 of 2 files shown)


wxmerkt:
Is this with ``libompl-dev``?

christian-rauch:
No, it's with `ros-melodic-ompl`. I haven't tried with `libompl-dev`, but I think that catkin will prefer the ROS libraries anyway.

wxmerkt:
Thank you very much for the fix :)

PR #377, opened 2018-09-26
operator for Hessian difference

christian-rauch:
On newer Eigen versions (>=3.3) I get the error: `error: ambiguous overload for ‘operator-’` for difference of Hessians. This is probably caused by additional new subtraction operators in Eigen.
Full log: [exotica_eigen_error.txt](https://github.com/ipab-slmc/exotica/files/2421335/exotica_eigen_error.txt)

A workaround for now is the spcialisation of the Hessian difference.
It's not a proper solution, but I could not find an alternative solution that makes it explicite to Eigen which operator to use.

### DIFF
--- a/examples/exotica_examples/tests/test_problems.cpp
+++ b/examples/exotica_examples/tests/test_problems.cpp
@@ -7,6 +7,24 @@ using namespace exotica;
 #define CREATE_PROBLEM(X, I) std::shared_ptr<X> problem = createProblem<X>(#X, I);
 #define NUM_TRIALS 100
 
+// workaround for "error: ambiguous overload for ‘operator-’" that appear in newer Eigen versions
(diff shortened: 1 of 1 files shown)


wxmerkt:
Which version of Eigen are you using on which operating system and compiler version? 3.3-beta/3.2.92 as shipped with Kinetic works fine.

christian-rauch:
I am on Ubuntu 18.04 / ROS melodic now.
Eigen: `libeigen3-dev/bionic,bionic,now 3.3.4-4 all`

VladimirIvan:
Can you rename the minuend and subtrahend to A, B and the difference to ret or something more generic. While the naming on its own would be correct, if we were to overload more operators each one of them would have differently named arguments. Also, those words are a mouthful.

PR #378, opened 2018-09-28
use system TinyXML2 library

christian-rauch:
This PR replaces the `tinyxml2` source inside exotica with the system version of tinyxml2 to prevent runtime crashes.

In general, I would recommend not to mix system libraries with project source files with the same API. This usually ends up with incompatible ABIs and makes tracking SIGSEGV very difficult.


christian-rauch:
It won't build on Ubuntu 14.04 / ROS indigo this way. The package `ros-indigo-cmake-modules` only includes `FindTinyXML.cmake`, not `FindTinyXML2.cmake`.
Is anyone still using ROS indigo? Can we switch the CI to ROS melodic?

Or is there another way we can make CMake discover tintxml2, which is indeed available on Ubuntu 14.04?

wxmerkt:
You can't change the current CI away from Indigo as Travis does not yet offer Xenial. I have a ROS Industrial CI under preparation which will support Kinetic and Melodic, but that is not ready yet (we need to change some things about our build and requirement system and how we run unit tests). Maybe try a pkg-config workaround and if it fails use the CMake one. There needs to be some way to do it with both, or we wait with replacing this until we can move CI up (maybe in one or two months once we find free cycles to get the Industrial CI to be compatible).

christian-rauch:
I updated the CI to `xenial`/`kinetic`. It builds now.

PR #381, opened 2018-09-28
ROS server publisher, node handle, string vector for pyexotica

christian-rauch:
This PR collects some commits that I used to interface with the `Server`.

- reduce code by using perfect forwarding for ROS publisher instantiation
- make `getNodeHandle` static, since Server is singleton anyway
- add support for `std::vector<std::string>`

### DIFF
--- a/exotica/include/exotica/Server.h
+++ b/exotica/include/exotica/Server.h
@@ -143,22 +143,16 @@ public:
         }
     }
 
-    template <typename T>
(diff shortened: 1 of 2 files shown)


wxmerkt:
Can you please rebase on master? This failed due to #380.

christian-rauch:
CI fails with an unrelated error: `/home/travis/catkin_ws/src/exotica/exotations/task_maps/task_map/include/task_map/CoM.h:38:37: fatal error: task_map/CoMInitializer.h: No such file or directory`

PR #390, opened 2018-10-30
correct gradient for 0-clipped distance (distance to line start)

christian-rauch:
This PR uses a different P2L gradient when the Point-to-Line distance is computed to the start of a finite line. It effectively becomes a Point-to-Point distance.

### DIFF
--- a/exotations/task_maps/task_map/src/Point2Line.cpp
+++ b/exotations/task_maps/task_map/src/Point2Line.cpp
@@ -91,9 +91,18 @@ void Point2Line::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen
         // direction from point to line
         const Eigen::Vector3d dv = direction(p);
         phi.segment<3>(i * 3) = dv;
-        for (int j = 0; j < J.cols(); j++)
+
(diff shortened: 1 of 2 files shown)


exotations/task_maps/task_map/src/Point2Line.cpp
```diff
@@ -91,9 +91,18 @@ void Point2Line::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen
         // direction from point to line
         const Eigen::Vector3d dv = direction(p);
         phi.segment<3>(i * 3) = dv;
-        for (int j = 0; j < J.cols(); j++)
+
+        if ((dv + p - line_start).norm() < 1e-9)
```
wxmerkt: How did you pick ``1e-9``? And could you use the system defined precision limits here?

christian-rauch: I thought there is no default epsilon in EXOTica? (I asked about this a while ago.)

christian-rauch:
I am using the C++ standard epsilon and added come clang-format changes for exotica_python.

PR #397, opened 2018-11-12
Adding an example for EffAxisAlignment task map.

cmower:
* New example scripts for `EffAxisAlignment` task map.
* See [here](https://youtu.be/pQraW8rs3cw).

### DIFF
--- a/examples/exotica_examples/package.xml
+++ b/examples/exotica_examples/package.xml
@@ -18,6 +18,7 @@
   <depend>task_map</depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>rviz</exec_depend>
+  <exec_depend>joy</exec_depend>
   <test_depend>rostest</test_depend>
   <test_depend>val_description</test_depend>
 </package>
--- a/examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
+++ b/examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
@@ -14,11 +14,11 @@
     <PlanningScene>
       <Scene>
         <JointGroup>arm</JointGroup>
-	<URDF>{exotica_examples}/resources/robots/lwr_simplified.urdf</URDF>
-	<SRDF>{exotica_examples}/resources/robots/lwr_simplified.srdf</SRDF>
+        <URDF>{exotica_examples}/resources/robots/lwr_simplified.urdf</URDF>
+        <SRDF>{exotica_examples}/resources/robots/lwr_simplified.srdf</SRDF>
       </Scene>
     </PlanningScene>
-    
+
     <Maps>
       <EffPosition Name="Position">
         <EndEffector>
@@ -27,8 +27,8 @@
       </EffPosition>
 
       <EffAxisAlignment Name="Direction">
-	<EndEffector>
-	  <Frame Direction="0 1 0" Axis="0 0 1" Link="lwr_arm_6_link"/>
+        <EndEffector>
+          <Frame Direction="0 1 0" Axis="0 0 1" Link="lwr_arm_6_link"/>
         </EndEffector>
       </EffAxisAlignment>
 
@@ -38,9 +38,9 @@
       <Task Task="Position" Rho="1"/>
       <Task Task="Direction" Rho="1"/>
     </Cost>
-    
+
     <W> 7 6 5 4 3 2 1 </W>
-    
+
   </UnconstrainedEndPoseProblem>
 
 </ExampleConfig>
--- a/examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
+++ b/examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" ?>
+<ExampleConfig>
+
+  <IKsolver Name="MySolver">
+    <MaxIterations>100</MaxIterations>
+    <MaxStep>0.1</MaxStep>
+    <Tolerance>1e-5</Tolerance>
+    <Alpha>1.0</Alpha>
+    <C>1e-3</C>
+  </IKsolver>
+
+  <UnconstrainedEndPoseProblem Name="MyProblem">
+
+    <PlanningScene>
+      <Scene>
+        <JointGroup>arm</JointGroup>
+	<URDF>{exotica_examples}/resources/robots/lwr_simplified.urdf</URDF>
+	<SRDF>{exotica_examples}/resources/robots/lwr_simplified.srdf</SRDF>
+      </Scene>
+    </PlanningScene>
+    
+    <Maps>
+      <EffPosition Name="Position">
+        <EndEffector>
+          <Frame Link="lwr_arm_6_link" LinkOffset="0 0 0 0.7071 0  0.7071 0"/>
+        </EndEffector>
+      </EffPosition>
+
+      <EffAxisAlignment Name="Direction">
+	<EndEffector>
+	  <Frame Direction="0 1 0" Axis="0 0 1" Link="lwr_arm_6_link"/>
+        </EndEffector>
+      </EffAxisAlignment>
+
+    </Maps>
+
+    <Cost>
+      <Task Task="Position" Rho="1"/>
+      <Task Task="Direction" Rho="1"/>
+    </Cost>
+    
+    <W> 7 6 5 4 3 2 1 </W>
+    
+  </UnconstrainedEndPoseProblem>
+
+</ExampleConfig>
--- a/examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
+++ b/examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
@@ -30,7 +30,7 @@ class Example(object):
         
 
         # Setup EXOtica
-        self.solver=exo.Setup.loadSolver('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
+        self.solver = exo.Setup.loadSolver('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
         self.problem = self.solver.getProblem()
 
         # Setup ROS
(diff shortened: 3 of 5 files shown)


examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
```diff
@@ -0,0 +1,47 @@
...
+    <Alpha>1.0</Alpha>
+    <C>1e-3</C>
+  </IKsolver>
+
+  <UnconstrainedEndPoseProblem Name="MyProblem">
+
+    <PlanningScene>
+      <Scene>
+        <JointGroup>arm</JointGroup>
+	<URDF>{exotica_examples}/resources/robots/lwr_simplified.urdf</URDF>
```
wxmerkt: Indentation (use an XML formatter)

examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
```diff
@@ -0,0 +1,47 @@
...
+      </EffAxisAlignment>
+
+    </Maps>
+
+    <Cost>
+      <Task Task="Position" Rho="1"/>
+      <Task Task="Direction" Rho="1"/>
+    </Cost>
+
+    <StartState>0 0 0 0 0 0 0</StartState>
```
wxmerkt: If 0s, not required.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+from numpy import array
+from numpy import matrix
+import math
+from pyexotica.publish_trajectory import *
+from time import sleep
+import task_map_py
+from sensor_msgs.msg import Joy, JointState
+from std_msgs.msg import Float64MultiArray as FloatArray
+
+rospy.init_node('example_ik_eff_axis_alignment_node')
```
wxmerkt: Move to ``if __name__ == "__main__"``- not the top of the file.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self):
+
+        exo.Setup.initRos()
+        (sol, prob)=exo.Initializers.loadXMLFull('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
+        self.problem = exo.Setup.createProblem(prob)
+        self.solver = exo.Setup.createSolver(sol)
+        self.solver.specifyProblem(self.problem)
+        self.scene = self.problem.getScene()
+        
+        self.q=array([0.0]*7)
```
wxmerkt: Python formatting - e.g. PEP8 without the linebreaks recommended (inconsistent).

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+LIMITS = array([XLIM,\
+                YLIM,\
+                ZLIM])
+
+class Example(object):
+
+    def __init__(self):
+
+        exo.Setup.initRos()
+        (sol, prob)=exo.Initializers.loadXMLFull('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
```
wxmerkt: Unless necessary, use the new initialisation with ``solver = exo.Setup.loadSolver({}); problem = solver.getProblem()``

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+        self.problem = exo.Setup.createProblem(prob)
+        self.solver = exo.Setup.createSolver(sol)
+        self.solver.specifyProblem(self.problem)
+        self.scene = self.problem.getScene()
+        
+        self.q=array([0.0]*7)
+        signal.signal(signal.SIGINT, sigIntHandler)
```
wxmerkt: Signal maybe should be in the "main" block rather than here.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python
+import rospy
+import pyexotica as exo
+import numpy as np
+from numpy import array
+from numpy import matrix
```
wxmerkt: You are not using matrix, hence, remove. In general it is discouraged to use numpy.matrix. Also you can group array under np, i.e., change code from array to np.array (or only include array if you are not using anything else from numpy).

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+import math
```
wxmerkt: Unused, remove.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+from pyexotica.publish_trajectory import *
```
wxmerkt: Maybe just ``publishPose`` and ``sigIntHandler`` - we currently have hidden dependencies that are getting into examples by using the ``*``. Once we fix the ``publish_trajectory`` those examples would be broken.

examples/exotica_examples/package.xml
```diff
@@ -17,5 +17,6 @@
   <depend>ik_solver</depend>
   <depend>task_map</depend>
   <depend>robot_state_publisher</depend>
+  <depend>joy</depend>
```
wxmerkt: Consider making ``run_depend``.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
cmower: Is there a specific reason for this? Or is it just for style?

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
cmower: This comes from copy/paste. Perhaps `from numpy import matrix` is used in another example file. Either way, this line will be removed.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
wxmerkt: It's wrong - when you like to include the file it can cause crashes and hangs otherwise (e.g., multiplicate initialisations). You should _not_ initialise a rospy node unless when needed in the main part. While this technically is a standalone example, since it contains a class it may be used differently or included into a notebook or larger script. It is a common mistake when people pick up this habit and has lead to a few users reporting "stuff hangs/doesnt work".

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
wxmerkt: I understand, a lot of the old examples contain garbage for historical reasons. We should clean them up and update API at some point.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
cmower: This is fine but what about `signal`? Should this be imported as an additional module or from `publish_trajectory`? I.e. 1 or 2 as follows.

1:

```
from pyexotica.publish_trajectory import publishPose, sigIntHandler, signal
```

2:

```
import signal
from pyexotica.publish_trajectory import publishPose, sigIntHandler
```

examples/exotica_examples/package.xml
cmower: We use format 2 for `package.xml` so need `exec_depend` rather than `run_depend`.

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
wxmerkt: (2) is the better option.

examples/exotica_examples/package.xml
wxmerkt: Yes, I realised that later - sorry, please use ``exec_depend``.

examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
```diff
@@ -38,10 +38,9 @@
       <Task Task="Position" Rho="1"/>
       <Task Task="Direction" Rho="1"/>
     </Cost>
-
-    <StartState>0 0 0 0 0 0 0</StartState>
-    <NominalState>0 0 0 0 0 0 0</NominalState>
```
wxmerkt: Sorry I should have been more clear: 

1. You can remove StartState if it is all zeros. That will be the default.
2. If you remove NominalState, using null space will be _turned off_. Setting NominalState to zero will activate null space for solvers that support it (e.g., pseudo-inverse IK). As such removing L43 changed the functionality of this demo [not sure whether that was intended].

examples/exotica_examples/scripts/example_ik_eff_axis_alignment.py
```diff
@@ -20,23 +18,20 @@
...
 class Example(object):
 
     def __init__(self):
 
         exo.Setup.initRos()
-        (sol, prob)=exo.Initializers.loadXMLFull('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
-        self.problem = exo.Setup.createProblem(prob)
-        self.solver = exo.Setup.createSolver(sol)
-        self.solver.specifyProblem(self.problem)
+        self.solver=exo.Setup.loadSolver('{exotica_examples}/resources/configs/example_eff_axis_alignment.xml')
```
wxmerkt: Inconsistent formatting remains (spaces around ``=``).

examples/exotica_examples/resources/configs/example_eff_axis_alignment.xml
cmower: I don't think it makes much difference having `<NominalState>` there. The goal of the example script, I think, should to be to highlight how `EffAxisAlignment` can be used and how the task map specifically works. I feel a general explanation of `<NominalState>` should exist elsewhere in the documentation.

Issue #398, opened 2018-11-12
Build failing on ros melodic, fcl_catkin requires manual install. 


cmower:
**Problem**
When building on Ubuntu 18.04 using ROS Melodic the following error appears. 
```
Errors     << collision_scene_fcl_latest:cmake /home/chris/Projects/University/kuka_ws/logs/collision_scene_fcl_latest/build.cmake.000.log
CMake Error at /opt/ros/melodic/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "fcl_catkin" with
  any of the following names:

    fcl_catkinConfig.cmake
    fcl_catkin-config.cmake

  Add the installation prefix of "fcl_catkin" to CMAKE_PREFIX_PATH or set
  "fcl_catkin_DIR" to a directory containing one of the above files.  If
  "fcl_catkin" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  CMakeLists.txt:7 (find_package)
```
**Workaround**
`$ sudo apt-get install ros-melodic-fcl-catkin`
or
pull [`https://github.com/ipab-slmc/fcl_catkin.git`](https://github.com/ipab-slmc/fcl_catkin.git) into workspace

wxmerkt:
rosdep should have resolved this and it is unclear why it didnt.

VladimirIvan:
I could not reproduce this on fresh install of 16.04 with melodic. `rosdep` installs the package correctly. Only `pybind11_catkin` has to be compiled from source. Did you do `rosdep update`?

PR #399, opened 2018-11-12
Temporary workaround for fcl catkin issue #398

cmower:


wxmerkt:
I am hesitant on adding this until I get more confirmation - I installed a virtual machine myself to confirm and cannot reproduce this behaviour. The dependency is clearly listed (see [here](https://github.com/ipab-slmc/exotica/blob/master/exotations/collision_scene_fcl_latest/package.xml#L13)) and should be successfully resolved by rosdep.

PR #411, opened 2018-11-26
Taskmap: backward finite differencing support

wxmerkt:
Computes Taskmap Jacobians by backward finite differencing if no explicit Jacobian is defined. The default behaviour is now as follows for Taskmaps:

- 0th-order update _needs_ to be defined (``update(x, phi)``).
- 1st-order update _should_ be defined where possible (``update(x, phi, J)``; do not avoid to do chain rule or use the corresponding functions that can compute Jdot for instance (``KinematicTree::ComputeJdot``)) - if it is not defined, the default behaviour is to compute J by backward finite differencing. [This is very, very expensive and will slow things down quite a bit]
- 2nd-order updates can be specified if desired - where they are not, the Hessian will be approximates as ``J^T * J``.

cc: @cmower

### DIFF
--- a/exotica/src/TaskMap.cpp
+++ b/exotica/src/TaskMap.cpp
@@ -77,13 +77,6 @@ void TaskMap::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::M
     Eigen::VectorXd x_original(x);
     update(x_original, phi);
 
-    // Do not run finite differencing if phi is zero.
-    if (phi.isApprox(Eigen::VectorXd::Zero(taskSpaceDim())))
-    {
-        J.setZero();
-        return;
-    }
-
     // Backward finite differencing.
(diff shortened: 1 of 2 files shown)


exotica/src/TaskMap.cpp
```diff
@@ -68,9 +68,36 @@ std::vector<KinematicFrameRequest> TaskMap::GetFrames()
...
+    // Store original x (needs to be reset later).
+    Eigen::VectorXd x_original(x);
+    update(x_original, phi);
+
+    // Do not run finite differencing if phi is zero.
+    if (phi.isApprox(Eigen::VectorXd::Zero(taskSpaceDim())))
+    {
+        J.setZero();
+        return;
+    }
```
VladimirIvan: This means that, e.g., if the end-effector is at its goal J=0. That is not true!

PR #413, opened 2018-11-28
Added CollisionSphere example

### LINKED ISSUE #396: Planning example for AICO with collision avoidance
Hi, I am trying to use AICO to plan a trajectory for LWR while avoiding collision. I tried to follow the examples provided, but the only ones that I can find are for the time indexed sampling problem. 

In the documentation I can only find Sphere and SphereCollision for task maps to check for collision. Does it mean that I have to define a sphere for each link and object, and then define a SphereCollision map for each pair? And if yes, how should I determine the location and the size of the spheres?

It would be very helpful if there is one working example using AICO and collision avoidance.

Thank you!
@VladimirIvan

VladimirIvan:
Fixes #396 
@teguhSL

### DIFF
--- a/examples/exotica_examples/scripts/example_sphere_collision.py
+++ b/examples/exotica_examples/scripts/example_sphere_collision.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+
+import pyexotica as exo
+from pyexotica.publish_trajectory import *
+
+# Custom pose and trajectory publishing to force update the task map debug marker publishing
+def publishPose(q, problem, t=0):
+    problem.update(q, t)
+    problem.getScene().getSolver().publishFrames()
+
+
+def publishTrajectory(traj, T, problem):
+    if len(traj) == 0:
+        print("Trajectory has zero elements")
+        raise
+    signal.signal(signal.SIGINT, sigIntHandler)
+    print('Playing back trajectory '+str(T)+'s')
+    dt = float(T)/float(len(traj))
+    t = 0
+    while True:
+        try:
+            publishPose(traj[t], problem, t)
+            sleep(dt)
+            t = (t+1) % len(traj)
+        except KeyboardInterrupt:
+            return False
+
+
+exo.Setup.initRos()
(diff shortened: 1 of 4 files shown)


examples/exotica_examples/scripts/example_sphere_collision.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+import pyexotica as exo
+from pyexotica.publish_trajectory import *
+
+# Custom pose and trajectory publishing to force update the task map debug marker publishing
+def publishPose(q, problem, t=0):
```
wxmerkt: I think it may make sense to make this the default behaviour for the regular publishing?

examples/exotica_examples/scripts/example_sphere_collision.py
```diff
@@ -0,0 +1,38 @@
...
+            t = (t+1) % len(traj)
+        except KeyboardInterrupt:
+            return False
+
+
+exo.Setup.initRos()
+solver = exo.Setup.loadSolver('{exotica_examples}/resources/configs/sphere_collision.xml')
+problem = solver.getProblem()
+
+for t in range(0, problem.T-1):
```
wxmerkt: Set Rho=0 in XML and then only set ``problem.setRho('Pose', 1, -1)`` <-- the -1 will automatically default to the last timestep.

examples/exotica_examples/scripts/example_sphere_collision.py
VladimirIvan: It requires integer indexing (unlike real valued time value) for the time indexed problem update. This is incompatible with publishing the stream of poses from IK.
We can declare a separate `publishTimeIndexedPose` and use that instead though.

examples/exotica_examples/scripts/example_sphere_collision.py
VladimirIvan: I specifically wanted to set the Rho value in XML rather than in python. That was you can see both values and tune them from one place.

Issue #412, opened 2018-11-27
Eigen Alignment Issues


wxmerkt:
We are currently having severe Eigen alignment issues, at least on Ubuntu 16.04 with gcc (which is our de facto default setup). The underlying reason here is that 

(a) Ubuntu 16.04 ships a beta version of Eigen 3.3 by default.
(b) The default compiler is gcc 5.4.0.
(c) The AVX alignment here is broken (random segmentation faults).

**TODO:**

- Check our alignment and allocation operators.

Good resources:

- http://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html 
- https://github.com/RobotLocomotion/drake/issues/1854

VladimirIvan:
Is this different from #315 ?

wxmerkt:
It appears to be. #315 can be triggered with solely Exotica core and is related already to the 3.3-beta and could be fixed with workarounds. 

This one here is more explicitly to track required work to fix our allocators/alignments (which hopefully should close both at once).

Issue #315, opened 2018-06-05
Default 16.04 Eigen breaks Eigen memory mapping


wxmerkt:
Ubuntu 16.04 with ``libeigen3-dev 3.3~beta1-2``, gcc 5.4.0

Noticed with BayesianIK crashing in line 359.

wxmerkt:
This issue sadly reoccurred. It can be triggered by using ``setZero`` or ``setOnes`` on non-fixed-size (!) vectors using the default Eigen on 16.04.

- [Eigen upstream bug report](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1427)

PR #415, opened 2018-11-29
Pass by const reference instead of pass by value, change from Affine to Isometry transforms

### LINKED ISSUE #414: Unify all transforms to use Eigen::Isometry unless explicitly required
Internally we extensively use ``Eigen::Affine3d``, which allows for affine transformations including shear and scaling. For rigid body transformations, it uses the affine inversion by default which runs SVD and is thus much more expensive to compute (unless specified (``.inverse(Eigen::Isometry)``).

Thus, I'd propose to check where we can replace Affine3d with Isometry3d.

wxmerkt:
A bunch of fixes from debugging the memory issue which are confirmed to work on different machines with the various compilation settings (release, debug, optimisation, platform targeting). My original issue appears to come down to a defective/broken CPU in my workstation [works just fine on two other machines of same generation/model and software setup].

- Moves from pass by value to pass by const reference where possible in problems
- Cleans up some AICO computation and removes typo
- Moves from Affine3d to Isometry3d (resolves #414)
- Removed dead code from old initialisation
- Removes ownership of RobotModel in Scene - now only in KinematicTree and implicitly in the planning scene inside the scene

### DIFF
--- a/exotations/solvers/aico/include/aico/AICOsolver.h
+++ b/exotations/solvers/aico/include/aico/AICOsolver.h
@@ -103,7 +103,7 @@ public:
        */
     void saveCosts(std::string file_name);
 
-    void getStats(std::vector<SinglePassMeanCoviariance>& q_stat_);
+    void getStats(std::vector<SinglePassMeanCovariance>& q_stat_);
 
 protected:
     /** \brief Initializes message data.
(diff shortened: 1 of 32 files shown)


VladimirIvan:
Are #315 and #412 still unresolved?

wxmerkt:
I will treat this for now as "likely hardware issues" since it solely happens on a single machine. We still have not properly done alignments - but it appears to work for now so I have closed both issues.

PR #416, opened 2018-11-30
Fix quaternion conversion

### LINKED ISSUE #334: SNOPT Derivative Checking throws on our orientation derivatives
When turning derivative verification in SNOPT on it fails if using any of our orientation task maps. Using any other task map works. Our own finite differencing tests pass, but fail when increasing the ``eps`` even on position (which passes with SNOPT).

VladimirIvan:
- Fixes inaccuracy in Quaternion conversion.
- Fixes task map tests (no longer a special case case for orientation).
- Fixes #334

### DIFF
--- a/exotations/collision_scene_fcl_latest/CMakeLists.txt
+++ b/exotations/collision_scene_fcl_latest/CMakeLists.txt
@@ -3,14 +3,11 @@ project(collision_scene_fcl_latest)
 
 add_compile_options(-std=c++11)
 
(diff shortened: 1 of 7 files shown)


wxmerkt:
Thank you very much - this will help a lot with our numerical issues :-)

wxmerkt:
The CI failed due to (a) formatting and (b) tests failing (1.02x*1e-5) on Melodic

PR #419, opened 2018-12-10
Upgrade clang format requirement to 3.9

wxmerkt:
Updates our requirement for the clang-format version to **3.9** as 3.8 is no longer available on 18.04.

cc: @christian-rauch


wxmerkt:
This pull request now includes a fix for the upcoming API breaking change from https://github.com/ros-planning/moveit/pull/1096 - i.e., that on Melodic the planning scene interface requires Isometry3d as well (cc @christian-rauch).

christian-rauch:
Thanks for pointing to the moveit repo.
Does the CI use a newer version of the moveit packages? The error (`YOU_PERFORMED_AN_INVALID_TRANSFORMATION_CONVERSION`) actually only appeares on the CI build server.
On my setup (moveit 0.10.5), /opt/ros/melodic/include/moveit/planning_scene/planning_scene.h:676 still cotnains the old API: `bool loadGeometryFromStream(std::istream& in, const Eigen::Affine3d& offset);`.

christian-rauch:
I am actually baffled that they chose to break the API in a stable long term release...

wxmerkt:
The way ROS packages are released they are first put on a SHADOW_FIXED server (which CI uses by default) before they are "synced" to the main debian packages everybody gets. My machine is also still on the 0.10.5 until the next sync, but the CI will always use what SHADOW_FIXED provides. We could opt to run both released and shadow repositories to make tracing these down easier but I had a clear suspicion in this case. [We are probably one of the few users of the deep down MoveIt APIs]

PR #431, opened 2018-12-11
Set fixed joints

christian-rauch:
This PR implements functions to fix model joints during the optimisation. This is implemented by setting `IsControlled` to false for a given set of links or joints.

An example use for fixing the 6D pose of the robot base:
```Python
self.problem_conf = ('exotica/UnconstrainedEndPoseProblem', {...})
self.problem = exo.Setup.createProblem(self.problem_conf)
# fix 6D base pose
self.problem.getScene().getSolver().setLinksFixed(set(self.problem.getScene().getJointNames()[:6]))
```

### DIFF
--- a/exotica/include/exotica/KinematicTree.h
+++ b/exotica/include/exotica/KinematicTree.h
@@ -212,6 +212,7 @@ public:
     std::vector<std::shared_ptr<KinematicElement>> getModelTree() { return ModelTree; }
     std::map<std::string, std::weak_ptr<KinematicElement>> getTreeMap() { return TreeMap; }
     std::map<std::string, std::weak_ptr<KinematicElement>> getCollisionTreeMap() { return CollisionTreeMap; }
+    std::map<std::string, std::weak_ptr<KinematicElement>>& getModelJointsMap() { return ModelJointsMap; }
     bool doesLinkWithNameExist(std::string name);  //!< Checks whether a link with this name exists in any of the trees
     bool Debug = false;
 
--- a/exotica_python/src/Exotica.cpp
+++ b/exotica_python/src/Exotica.cpp
@@ -189,9 +189,9 @@ void setJointsControlled(KinematicTree& tree, const std::set<std::string>& joint
 
 void setLinksControlled(KinematicTree& tree, const std::set<std::string>& links, const bool negate = false)
 {
-    for (auto & [ name, joint ] : tree.getModelJointsMap())
+    for (auto & [ name, link ] : tree.getTreeMap())
     {
-        joint.lock()->IsControlled = (joint.lock()->ControlId != -1) && (negate ^ bool(links.count(joint.lock()->Segment.getName())));
+        link.lock()->IsControlled = (link.lock()->ControlId != -1) && (negate ^ bool(links.count(name)));
     }
 }
 
--- a/exotica_python/src/Exotica.cpp
+++ b/exotica_python/src/Exotica.cpp
@@ -179,6 +179,14 @@ Eigen::MatrixXd Solve(std::shared_ptr<MotionSolver> sol)
     return ret;
 }
 
+void setJointsFixed(KinematicTree& tree, const std::set<std::string> joints)
+{
+    for (auto & [ name, joint ] : tree.getModelJointsMap())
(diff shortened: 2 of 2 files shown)


christian-rauch:
The hash map iteration (`for(auto & [key,value] : map) { ... }`) requires C++17. I assume you are not willing to bump EXOTica's requirements to this level.
Let me know if you are willing to accept the implemented feature and I will implement a C++11 equivalent.

wxmerkt:
I dont think we want to merge this. 
We currently support this scenario via the SRDF: You can set what the controlled joints are there (via the planning group). If you want to have a different set of joints, you select/set a different planning group. I have the feeling there are further dependent changes required down the line or implementing this change as requested (beyond the C++17 requirement) could introduce bugs we can't yet easily anticipate (the `IsControlled` is rather central).

christian-rauch:
I also use this functionality to optimise the kinematic chain link-wise, e.g. only a certain link/joint for a couple of iterations. And this is not really possible using a static SRDF file.
While you can achieve something similar with SRDF settings, it will require many different joint groups that need to be manually created for every robot and every joint set. You also need to create the scene every time you change the optimised joint group.

VladimirIvan:
So what is the behaviour here?
1. Fixed joints stay at the last position they've been updated to?
1. Do you still set them with `setModelState`?
1. Does the list of controlled joints get updated propagated to the rest of the problem? ie, does this change the number of columns of the Jacobian and the size of the state vector?

christian-rauch:
1. Yes. But you could imagine that you at any point in time only want to optimise a subset of joints.
2. I am not using `setModelState`. But I set the starting state from the previous solution.
3. The Jacobian and robot state dimension stays the same.

I am propagating the set of tracked joints to my task, to cancle phi for not-tracked links.

If I am not wrong, not controlled joints (`IsControlled`=false) are seen as fixed joints (i.e. as if they would have been declared `fixed` in the URDF)?

christian-rauch:
Apart from setting the joint limits of a joint to its current value to prevent its update, what other options are there to optimise an arbitrary set of joints that is updated at runtime?

VladimirIvan:
> If I am not wrong, not controlled joints (`IsControlled`=false) are seen as fixed joints (i.e. as if they would have been declared `fixed` in the URDF)?

If you declare a joint as fixed in the URDF, the controlled joints list and the Jacobian won't contain that joint at all. This will then get passed to the TaskMaps, Problem, Solver in a transparent way.

What you do will only set the flag to stop updating the Jacobian. TaskMaps, Problems and Solvers won't know about this. The result will be a cost that is the same as the one with the joint being controlled but the Jacobian will be incorrect. This will mess up 100% of the solvers if used as is.

> I am propagating the set of tracked joints to my task, to cancle phi for not-tracked links.

That is a hack that will fix your task but it will break all other tasks used in the same problem.

Only clear place where you could achieve this without breaking the rest of the infrastructure is in the solver. If the solver requests all joints to be updated but chooses to ignore a subset of them during some part of its computation, that is absolutely fine.That would be a very specific solver though. Probably bound to a robot model and the problem/task map.

christian-rauch:
> What you do will only set the flag to stop updating the Jacobian. TaskMaps, Problems and Solvers won't know about this. The result will be a cost that is the same as the one with the joint being controlled but the Jacobian will be incorrect. This will mess up 100% of the solvers if used as is.

Yes, one implication of this is that if you use a joint/frame that has been set as fixed/not-controlled in a task map, the taskmap will compute a cost for this task that cannot be minimised. But since the Jacobian columns for the non-controlled joints will be 0, this behaves as if the joints are fixed.

If I would define a cost on a link of the robot, that is set fixed in the URDF, wouldn't I get the same behaviour? Sure, the Jacobian would have a different dimension, but a solver would give the same q_d or not?

wxmerkt:
I had another conversation with Vlad - and setting links fixed link this breaks actually more than it helps: task maps need to be specialised etc. The correct solution would be to leave problems / kinematics / taskmaps untouched and in the solver have the option to freeze optimisation variables - and then there - ignore columns of the Jacobian. The solution as presented in this pull request breaks things :-(.

PR #435, opened 2018-12-12
obtain joint limits after task initialisation

christian-rauch:
The joint limits get reset to their defaults when calling `specifyProblem(<problem>)`. Since the task maps are also initialised during this, there is no way to change the joint limits for the `JointLimit` task, which fixes the joint limits at initialisation.

This PR updates the joint limits when `JointLimit::update` is called, hence allowing to change the joints limits from defaults, between the setup (`specifyProblem`) and the optimisation (`solve`).

### DIFF
--- a/exotations/solvers/exotica_ompl_solver/include/exotica_ompl_solver/ompl_solver.h
+++ b/exotations/solvers/exotica_ompl_solver/include/exotica_ompl_solver/ompl_solver.h
@@ -90,7 +90,7 @@ protected:
     void SetGoalState(Eigen::VectorXdRefConst qT, const double eps = 0);
     void PreSolve();
     void PostSolve();
-    void GetPath(Eigen::MatrixXdRef traj, ompl::base::PlannerTerminationCondition &ptc);
+    void GetPath(Eigen::MatrixXd &traj, ompl::base::PlannerTerminationCondition &ptc);
     OMPLSolverInitializer init_;
     std::shared_ptr<ProblemType> prob_;
     ompl::geometric::SimpleSetupPtr ompl_simple_setup_;
--- a/exotations/task_maps/task_map/include/task_map/JointLimit.h
+++ b/exotations/task_maps/task_map/include/task_map/JointLimit.h
@@ -65,7 +65,7 @@ private:
     Eigen::VectorXd low_limits_;   //	Lower joint limits
     Eigen::VectorXd high_limits_;  //	Higher joint limits
     Eigen::VectorXd tau_;          //	Joint limits tolerance
-    double percent;
+    double safe_percentage_;
     JointLimitInitializer init_;
     int N;
 };
--- a/exotations/task_maps/task_map/include/task_map/JointLimit.h
+++ b/exotations/task_maps/task_map/include/task_map/JointLimit.h
@@ -62,9 +62,6 @@ public:
     virtual int taskSpaceDim();
 
 private:
-    Eigen::VectorXd low_limits_;   //	Lower joint limits
-    Eigen::VectorXd high_limits_;  //	Higher joint limits
-    Eigen::VectorXd tau_;          //	Joint limits tolerance
     double safe_percentage_;
     JointLimitInitializer init_;
     int N;
--- a/exotations/task_maps/task_map/include/task_map/JointLimit.h
+++ b/exotations/task_maps/task_map/include/task_map/JointLimit.h
@@ -65,6 +65,7 @@ private:
     Eigen::VectorXd low_limits_;   //	Lower joint limits
     Eigen::VectorXd high_limits_;  //	Higher joint limits
     Eigen::VectorXd tau_;          //	Joint limits tolerance
+    double percent;
     JointLimitInitializer init_;
     int N;
 };
--- a/exotations/task_maps/task_map/src/JointLimit.cpp
+++ b/exotations/task_maps/task_map/src/JointLimit.cpp
@@ -52,7 +52,7 @@ void JointLimit::assignScene(Scene_ptr scene)
 
 void JointLimit::Initialize()
 {
-    percent = init_.SafePercentage;
+    safe_percentage_ = init_.SafePercentage;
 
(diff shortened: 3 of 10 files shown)


exotations/task_maps/task_map/include/task_map/JointLimit.h
```diff
@@ -65,6 +65,7 @@ class JointLimit : public TaskMap, public Instantiable<JointLimitInitializer>
     Eigen::VectorXd low_limits_;   //	Lower joint limits
     Eigen::VectorXd high_limits_;  //	Higher joint limits
     Eigen::VectorXd tau_;          //	Joint limits tolerance
+    double percent;
```
wxmerkt: `safe_percentage_`

exotations/task_maps/task_map/src/JointLimit.cpp
```diff
@@ -79,6 +70,16 @@ void JointLimit::update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi)
 {
     if (phi.rows() != N) throw_named("Wrong size of phi!");
     phi.setZero();
+
+    const Eigen::MatrixXd limits = scene_->getSolver().getJointLimits();
+
+    for (int i = 0; i < N; i++)
```
wxmerkt: Couple of changes required:
(a) `low_limits_` and `high_limits_` and `tau_` do not strictly have to be member variables (but perhaps good for no alloations)
(b) Replace with (vector operations)

```cpp
Eigen::VectorXdRefConst low_limits = limits.col(0);
Eigen::VectorXdRefConst high_limits = limits.col(1);
tau_ = 0.5 * safe_percentage_ * (high_limits - low_limits);
```

christian-rauch:
If you don't mind, I replaced the hard coded specific joint limits for the floating base with unbound joint limits. From now on, a user needs to manually define these joints limits.

exotica/src/KinematicTree.cpp
```diff
@@ -907,20 +907,21 @@ void KinematicTree::resetJointLimits()
 
     ///	Manually defined floating base limits
     ///	Should be done more systematically with robot model class
+    const double inf = std::numeric_limits<double>::infinity();
```
wxmerkt: Can you use a `constexpr` here? And also use a definition for pi: `constexpr double pi() { return std::atan(1)*4; }`

wxmerkt: Infinity may actually not play nicely with sampling-based planners from OMPL, I realised.

christian-rauch: > Infinity may actually not play nicely with sampling-based planners from OMPL, I realised.

Right :-) But does it recognise user-chosen joint limits? In this case, a user has to select translation limits before sampling from bounds.

wxmerkt: We have exposed ``setFloatingBaseLimits`` so that could work _if_ a user remembers to do so. I'll have a look at this PR later again when CI passed (but that may require some formatting).

christian-rauch: > constexpr double pi() { return std::atan(1)*4; }

Why not use `M_PI`? I understand that `constexpr` is an inline hint. Since `M_PI` is a macro, it will be inline and not changeable.

wxmerkt: Cf. e.g. [here](https://meetingcpp.com/blog/items/cpp-and-pi.html)

christian-rauch: Alright. Do we have a place in exotica (maybe a "math_defines" header?) which collects all these constants?

christian-rauch: I placed both as `constexpr` just in the function.

wxmerkt:
I just realised, `infinity` is not an option for the bounds. It won't work with OMPL and neither with some of our optimisation solvers (e.g. in Snopt we use `1e20` for `no limit`). Do the OMPL freebase examples still work, or can you add one if none exists?

exotations/solvers/exotica_ompl_solver/include/exotica_ompl_solver/ompl_solver.h
```diff
@@ -90,7 +90,7 @@ class OMPLSolver : public MotionSolver
     void SetGoalState(Eigen::VectorXdRefConst qT, const double eps = 0);
     void PreSolve();
     void PostSolve();
-    void GetPath(Eigen::MatrixXdRef traj, ompl::base::PlannerTerminationCondition &ptc);
+    void GetPath(Eigen::MatrixXd traj, ompl::base::PlannerTerminationCondition &ptc);
```
wxmerkt: This needs to be a reference `&`

christian-rauch: ups

christian-rauch:
I am testing with the new `example_ompl_freebase_noros.py` (with Vlad's fix) which is an adaptation of `example_ompl.py` with the freebase config (`ompl_solver_demo_freebase.xml`).

Even with +/-10m bounds on the x-y planar position, the solver doesn't find a solution: `No solution found after 60.000018 seconds`.

christian-rauch:
The `example_ompl_freebase_noros.py` on master provides a solution. Limiting the translation to +/- 1e20 meter also provides a solution.
It appears as if OMPL does not respect the limits that are set in the python script.

wxmerkt:
That's because the solver appears to set up the bounds (and the OMPLxxStateSpace) during ``specifyProblem``. We should move that logic somewhere else, e.g., to `Solve()`.

christian-rauch:
I moved some stuff in `OMPLSolver` and `OMPLStateSpace` around and the example works now.

I am not familiar with OMPL. But it looks like I did the right thing :-)
- the new method `OMPLStateSpace::setBounds` does most of what the constructor used to do.
- some part of the setup in `OMPLSolver::specifyProblem` is now done in `OMPLSolver::Solve`

christian-rauch:
I added a check to make sure that the solver is only used with finite joint limits.

PR #440, opened 2018-12-13
Minor mod to smoothing example

cmower:
* `Rho=1e3` -> `Rho=1e2` for `Position` task map cost
* With new `Rho` value for `Position` task the benefit of the smoothing task maps is clearer when the example is launched
* Before, when `Rho=1e3`, the motion looks essentially the same as without the smoothing


wxmerkt:
> Before, when Rho=1e3, the motion looks essentially the same as without the smoothing

Not quite, if you went to the boundaries of reachable workspace you could see it.

Issue #391, opened 2018-10-31
Implementation of PseudoInverse


christian-rauch:
Which algorithm is the basis for `IKsolver::PseudoInverse`?

In the configuration J.cols() < J.rows() and C==0 the pseudo inverse is computed as:
```C++
Jpinv = inverseSymPosDef(J.transpose() * J + W) * J.transpose();
```

For  W=I (identity) is this supposed to provide the same result as the Moore-Penrose pseudoinverse?

I have a case where `IKsolver` gives me quite small update steps (q_d) for a sparse Jacobian, while Matlab's `pinv()` seems to provide the correct q_d.

christian-rauch:
Relacing 
```C++
Eigen::MatrixXd Jinv = PseudoInverse(prob_->Cost.S * prob_->Cost.J);
```
with 
```C++
Eigen::MatrixXd Jinv = (prob_->Cost.S*prob_->Cost.J).completeOrthogonalDecomposition().pseudoInverse();
```
already gives me a faster convergence.
How would I need to set `C` and `W` to obtain the same pseudo inverse from `IKsolver::PseudoInverse` as provided by Eigen and Matlab?

christian-rauch:
Quick update: I implemented the Gauss-Newton (actually Levenberg-Marquardt) solver https://github.com/christian-rauch/gauss_newton_exo_solver to compare the convergence on a synthetic case. In the case shown, I am adding links iteratively to the set of optimised links, but this should not matter. Optimisation for 5 x 5 iterations:

Inverse Kinematics (IK):
![exo_ik](https://user-images.githubusercontent.com/8226248/47926880-af875c80-deb9-11e8-91ca-dd62b763fe2b.gif)

Levenberg-Marquardt (LM):
![exo_lm](https://user-images.githubusercontent.com/8226248/47926895-bf06a580-deb9-11e8-82e5-a5886f316da0.gif)

The LM is using an adaptive damping factor with damping direction depending on the Hessian approximation as described here: https://uk.mathworks.com/help/optim/ug/least-squares-model-fitting-algorithms.html

VladimirIvan:
Thanks for pointing out the problems with the IK solver. I'll have a look at it and fix it.

I really like your minimal LM implementation. I'd love to add it to the core exotica.

PR #442, opened 2018-12-14
Levenberg-Marquardt Solver

christian-rauch:
Implementation of the Levenberg-Marquardt / Gauss-Newton method, based on https://uk.mathworks.com/help/optim/ug/least-squares-model-fitting-algorithms.html eq. 13.

I named the solver Levenberg-Marquardt since it has the additional adaptive damping option. However, in the default configuration, no damping is applied which then behaves like the Gauss-Newton.

I've observed (https://github.com/ipab-slmc/exotica/issues/391) that the Gauss-Newton converges faster but might be unstable close to the solution. This is why the damping needs to be set. The damping is adapted according to the decrease/increase of the error.


exotations/solvers/exotica_levenberg_marquardt_solver/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(exotica_lm_solver)
```
wxmerkt: Project name does not match the directory name. Perhaps change to ``exotica_levenberg_marquardt_solver`` as in the folder name (more descriptive)

exotations/solvers/exotica_levenberg_marquardt_solver/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+find_package(catkin REQUIRED COMPONENTS exotica)
+
+find_package(Eigen3 REQUIRED)
+
+AddInitializer(LMSolver)
```
wxmerkt: Change to more descriptive `LevenbergMarquardtSolver`

exotations/solvers/exotica_levenberg_marquardt_solver/package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>exotica_lm_solver</name>
+  <version>1.0.1</version>
+  <description>A Levenberg-Marquardt solver for EXOTica</description>
+
+  <maintainer email="Christian.Rauch@ed.ac.uk">Christian Rauch</maintainer>
+
+  <license>MIT</license>
```
wxmerkt: FYI: The rest of the repository is BSD licensed - both are compatible though so no action required.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
+#include <exotica/MotionSolver.h>
+#include <exotica/Problems/UnconstrainedEndPoseProblem.h>
+#include <exotica_lm_solver/LMSolverInitializer.h>
+
+namespace exotica
+{
+class LevenbergMarquardt : public MotionSolver, public Instantiable<LMSolverInitializer>
+{
+public:
+    virtual void Instantiate(LMSolverInitializer& init);
```
wxmerkt: Change to `void Instantiate(..) override;`

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    virtual void Solve(Eigen::MatrixXd& solution);
```
wxmerkt: Change to `void Solve(..) override;`

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    virtual void specifyProblem(PlanningProblem_ptr pointer);
```
wxmerkt: Change to `void specifyProblem(..) override;`

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+private:
+    LMSolverInitializer parameters_;
+
+    UnconstrainedEndPoseProblem_ptr prob_;  // Shared pointer to the planning problem.
+
+    double lambda = 0;  // damping factor
```
wxmerkt: `lambda_`

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+    if (pointer->type() != "exotica::UnconstrainedEndPoseProblem")
+    {
+        throw_named("This LevenbergMarquardt can't solve problem of type '" << pointer->type() << "'!");
+    }
+
+    MotionSolver::specifyProblem(pointer);
+
+    // generic problem
+    problem_ = pointer;
```
wxmerkt: Do you need both the generic and the specific problem?

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+        else if (parameters_.ScaleProblem == "Jacobian")
+        {
+            M = (J.transpose() * J).diagonal().asDiagonal();
+        }
+        else
+        {
+            throw std::runtime_error("no ScaleProblem of type " + parameters_.ScaleProblem);
+        }
+
+        qd = (J.transpose() * J + lambda * M).completeOrthogonalDecomposition().solve(J.transpose() * yd);
```
wxmerkt: `completeOrthogonalDecomposition` is sadly not available in 16.04's Eigen by default :'-(.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+    Eigen::VectorXd q = q0;
+    double error = std::numeric_limits<double>::infinity();
+    double error_prev = std::numeric_limits<double>::infinity();
+    Eigen::VectorXd yd;
+    Eigen::VectorXd qd;
+    for (size_t i = 0; i < getNumberOfMaxIterations(); iterations_ = ++i)
+    {
+        prob_->Update(q);
+
+        yd = prob_->Cost.S * prob_->Cost.ydiff;
```
wxmerkt: `Cost` may become a private member in the future.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
```diff
@@ -0,0 +1,138 @@
...
+    virtual void specifyProblem(PlanningProblem_ptr pointer);
+
+private:
+    LMSolverInitializer parameters_;
+
+    UnconstrainedEndPoseProblem_ptr prob_;  // Shared pointer to the planning problem.
+
+    double lambda = 0;  // damping factor
+
+    int iterations_ = -1;
```
wxmerkt: Technically unused, can be removed as a member.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
christian-rauch: That's a problem :-)

I used the LDLT decomposition before but got unstable results, i.e. different solutions (x) to the same problem (A,b). I haven't tried the LLT, but the instability with LDLT was really nasty and difficult to track down. I used sparse Jacobians back then and LDLT or LLT might work for other cases with less sparse Jacobians.
Because this issue is latent, I would generally recommend using the COD solver.

Options:
1. Do not enable/build on Ubuntu 16.04. Another good reason to upgrade to 18.04 :-)
2. Provide a Eigen catkin package with a sufficient high version. I don't really like this, but if someone really want to use this solver n 16.04.
3. Use a different library for solving. But this might result in similar issues as discussed with alternative decompositions.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
christian-rauch: Will it have a getter `getCost()`?

exotations/solvers/exotica_levenberg_marquardt_solver/package.xml
christian-rauch: I like the MIT because it's shorter :-)

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
wxmerkt: It will as needed - as evident by this solver - and `exotica_ik_solver`.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
wxmerkt: This is a tricky one:
(1) Easy to do with CMake, bad for user experience (people will ask 'why does this not work/is not available')
(2) May give us linking issues or runtime issues on release? Can we have it _exclusively_ as a private external for this project using `ExternalProject_Add` which is privately linked/not exposed?

christian-rauch: How do you add a private external dependency? CMake will fail for everyone that does not have access to the private repo.

I am in favour of option 1: Check the Eigen version and give a warning that this solver is not available for the poor souls that still use Ubuntu 16.04.

wxmerkt: I meant add the `ExternalProject_Add` to this project only. Do not use the `find_package(Eigen3)`. I am not 100% clear how to use the INTERFACE etc. features to avoid exposing Eigen-version-mismatch issues to other parts of the code, I am afraid.

> Check the Eigen version and give a warning that this solver is not available for the poor souls that still use Ubuntu 16.04.

Which, as of now, is every robot we have or are in the process of procuring. Upgrades to 18.04 can easily be 1-2 years out.

christian-rauch: I made the solver only built when Eigen version >= 3.3 is available. If someone wants to use this solver, he needs to manually install Eigen 3.3 or higher.

exotations/solvers/exotica_levenberg_marquardt_solver/CMakeLists.txt
```diff
@@ -23,3 +25,7 @@ install(TARGETS ${PROJECT_NAME}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 install(FILES exotica_plugins.xml DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+
+else()
+message("The Levenberg-Marquardt solver requires at least Eigen version 3.3.")
```
wxmerkt: Please make it a `WARNING` - otherwise it gets hidden by `catkin build -s`.

wxmerkt: Also, change to `if(NOT TARGET ...) message(WARNING...) return() endif()` - no else block required and can be removed later on much more easily.

christian-rauch: Nice, I didn't know that you can directly return from a CMake script file.

exotations/solvers/exotica_levenberg_marquardt_solver/src/LevenbergMarquardt.cpp
christian-rauch: Ubuntu 16.04 is using Eigen version 3.3 **beta1** (which has a serious bug: https://bugs.launchpad.net/ubuntu/+source/eigen3/+bug/1750794 you might want to be aware of).

The `CompleteOrthogonalDecomposition` was introduced in version 3.3 **beta2** with header `CompleteOrthogonalDecomposition.h`.
A fourth option would be to provide this header for users of version 3.3 beta1 (version 3.2.92 according to pkg-config and cmake).

exotations/solvers/exotica_levenberg_marquardt_solver/package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>exotica_levenberg_marquardt_solver</name>
+  <version>1.0.1</version>
```
wxmerkt: Are you okay if we change this to `5.0` before we run our binary release to melodic? Helps every package in this repository be at the same version level

wxmerkt:
Thank you very much for contributing the LM solver, it's a great addition :-)

exotations/solvers/exotica_levenberg_marquardt_solver/package.xml
christian-rauch: I wouldn't mind. Does every package of the exotica meta-package need to have the same version number?

Issue #444, opened 2018-12-14
OMPL state space locked (example_attach.py is broken)


christian-rauch:
The `example_attach.py` is broken (caused by my PR https://github.com/ipab-slmc/exotica/pull/435):
```
Traceback (most recent call last):
  File "[...]/exotica/examples/exotica_examples/scripts/example_attach.py", line 39, in <module>
    solution.append(ompl.solve().tolist())
RuntimeError: This state space is locked. No further components can be added
```

But the OMPL examples `example_ompl_noros.py` and `example_ompl_freebase_noros.py` are working.

christian-rauch:
The reason for this issue is that I moved the `lock()` to `setBounds`, which is called at every `Solve()`. Hence, the StateSpace is locked after the first Solve call.

Is it required to lock the OMPL state space?

PR #445, opened 2018-12-14
prevent change of OMPL state space after locking

christian-rauch:
This checks that the OMPL state space is not locked before attempting to change it during multipel calls to `Solve()`.
Fixes https://github.com/ipab-slmc/exotica/issues/444

### DIFF
--- a/examples/exotica_examples/CMakeLists.txt
+++ b/examples/exotica_examples/CMakeLists.txt
@@ -70,4 +70,6 @@ if(CATKIN_ENABLE_TESTING)
   add_dependencies(test_problems ${catkin_EXPORTED_TARGETS})
 
   add_rostest(test/python_tests.launch)
+
+  catkin_add_nosetests(test/run_tests.py)
 endif()
--- a/examples/exotica_examples/test/run_tests.py
+++ b/examples/exotica_examples/test/run_tests.py
@@ -1,0 +1,4 @@
+#!/usr/bin/env python
+import rosunit
+
+rosunit.unitrun('exotica_examples', 'ompl_solver_bounds', 'test.test_ompl_solver_bounds.OMPLLockedBoundsCase')
--- a/examples/exotica_examples/test/test_ompl_solver_bounds.py
+++ b/examples/exotica_examples/test/test_ompl_solver_bounds.py
@@ -1,0 +1,49 @@
+import unittest
+
+import pyexotica as exo
+
+class OMPLLockedBoundsCase(unittest.TestCase):
+
+    def test_lock_bounds(self):
+        # self.assertTrue(True)
+        ompl=exo.Setup.loadSolver('{exotica_examples}/resources/configs/example_manipulate_ompl.xml')
+
+        # set start and goal state
+        ompl.getProblem().startState = [1.5035205538438838, 0.8730168650583787, -1.6298590879018438, 1.7106630821349438, -0.8789956712153559, 0.1278222471656531, 0.0]
+        ompl.getProblem().goalState = [-1.5035205538442702, 0.8730168650583671, 1.6298590879018415, 1.7106630821349786, 0.8789956712153525, 0.12782224716566898, 0.0]
+
+        # 1st solve call
+        solution = None
+        try:
+            solution = ompl.solve()
+        except Exception:
+            pass
+        self.assertTrue(solution is not None)
+
+        # 2nd solve call
+        solution = None
+        try:
+            solution = ompl.solve()
+        except Exception:
+            pass
+        self.assertTrue(solution is not None)
+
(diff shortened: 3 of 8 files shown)


wxmerkt:
Thanks! I am not sure whether we need to lock it or not - either way, please add an else with a highlight to inform users that the state space is locked and bounds have not been updated. We could even check and only warn if the bounds are different from the bounds of the state space

christian-rauch:
This warning would then appear every time Solve is called.
Before moving this to solve, the majority of the setup was done in the constructor of the StateSpace and the initialiser of the solver. So basically, you couldn't change the bound previously either, before calling Solve.
I would prefer not to output information in this case. A user might call Solve many times with different start and goal states, and a warning would clutter the output.

Without locking, you would be able to change the bounds and the StateSpace for every call to Solve. I am not sure if there is an application for that and what implications, e.g. possible downsides, it will have.

christian-rauch:
I am not that familiar with OMPL to comment if or when the state space needs to be locked.

The documentation about `lock` states:
> Lock this state space. This means no further spaces can be added as components. This function can be for instance called from the constructor of a state space that inherits from CompoundStateSpace to prevent the user to add further components.

This is what we did previously (lock in constructor). Previously and now, a user would need to reconstruct the StateSpace with new bounds.

Are there examples/cases where the StateSpace does not get locked and is changed in between calls to `solve`?

wxmerkt:
I don't know. I'd recommend asking @yimingyanged - if in doubt, we can leave the state space unlocked, I think. We are not adding to it and we should simply check whether the size matches `problem_->N`.

christian-rauch:
If I remove the locking mechanism, the `example_attach.py` fails:
```
Traceback (most recent call last):
  File "./example_attach.py", line 39, in <module>
    solution.append(ompl.solve().tolist())
RuntimeError: The longest valid segment for state space RealVectorSpace2 must be positive.
Space settings:
Real vector state space 'RealVectorSpace2' of dimension 7 with bounds: 
  - min: 0 0 0 0 0 0 0 
  - max: 0 0 0 0 0 0 0 

Segmentation fault (core dumped)
```

christian-rauch:
So I am storing the original bounds when setting them first and check thereafter of they changed.

I tested by adding a small code snippet to the `example_attach.py` example inbetween calls to `solve`:
```Python
# Plan 3 trajectories
solution = []
for i in range(0,3):
    ompl.getProblem().startState = start_pose[i]
    ompl.getProblem().goalState = goal_pose[i]
    handleAttaching(i, start_pose[i], ompl.getProblem().getScene())
    solution.append(ompl.solve().tolist())
    # get and change current bounds
    jl = ompl.getProblem().getScene().getSolver().getJointLimits()
    jl[0] += 1
    ompl.getProblem().getScene().getSolver().setJointLimitsLower(jl[:,0])
    ompl.getProblem().getScene().getSolver().setJointLimitsUpper(jl[:,1])
    # consecutive calls to ompl.solve() will throw an error
```
and it correctly throws the error when calling `solve` for the second time.

Is there already a OMPL unit test, which I can extend to check of the exception is thrown?

wxmerkt:
Thanks for adding the check! We do not yet have OMPL solver unit tests. You could either add them to `exotica_examples`, where some of our tests that require resources are, or add new tests to the `exotica_ompl_solver` package (depending on apt-get-able resources; both a expected-failing and expected-succeeding case - you can fix the `RandomSeed` to produce a deterministic solution).

wxmerkt:
Found another issue related to this: #456

christian-rauch:
I added a test, as described in: http://wiki.ros.org/unittest#Loading_Tests_by_Name, using the same `example_manipulate_ompl` scene.

At the moment, the test is executed by running the test script manually: `./test/run_tests.py`. I am not sure where to add the run script, so that the test is automatically called with the others.

The `rostest test/python_tests.launch` already has 2 failed tests because of the missing `val_description` package.

wxmerkt:
Thanks, Christian. The tests are added in the `CMakeLists.txt`:
https://github.com/ipab-slmc/exotica/blob/1ac88790f32e90e2d564d25392b5b7c8fa22e9a6/examples/exotica_examples/CMakeLists.txt#L72

For testing on the build server, the `val_description` is provided by the `.rosinstall` file in the root of the directory (thus no failed tests).

christian-rauch:
I added the test via `catkin_add_nosetests(test/run_tests.py)`.

christian-rauch:
I reverted some of the changes in e612bbb9328b5a8beeee25711eeb2cd644738b53 to fix #459 .

VladimirIvan:
What would it take to recreate a state space that can have its bounds edited instead of throwing?

PR #448, opened 2018-12-17
Publish single list of markers with add and delete actions

### LINKED ISSUE #447: Published collision objects get deleted and added alternating
The collision scene object published via `publishPose(solution, ompl.getProblem())` alternative between `ADD` and `DELETEALL` causing flickering in RViz.
Instead of alternate publishing of the two actions, they should be added together in a single `MarkerArray` such that old markers are removed and new markers are added at the same time.

Note: If the IDs of marker objects do not change, there is no need to remove markers since old markers with the same ID can simply be overwritten.

christian-rauch:
This PR adds a global list of published markers to `KinematicTree`, including actions for `ADD` and `DELETEALL` (fixes #447 ).

- `resetModel` adds a `DELETEALL` marker to this global list, but does not publish them immediately
- the global list is published in `publishFrames` if the scene changed
- using enum `DELETEALL` (not available in indigo) for better readability and compatibiity with ROS message

### DIFF
--- a/exotica/include/exotica/KinematicTree.h
+++ b/exotica/include/exotica/KinematicTree.h
@@ -41,6 +41,7 @@
 
 #include <moveit/robot_model/robot_model.h>
 #include <tf_conversions/tf_kdl.h>
+#include <visualization_msgs/MarkerArray.h>
 #include <Eigen/Eigen>
 #include <kdl/jacobian.hpp>
 #include <kdl/tree.hpp>
@@ -271,6 +272,7 @@ private:
     std::vector<tf::StampedTransform> debugFrames;
     ros::Publisher shapes_pub_;
     bool debugSceneChanged;
+    visualization_msgs::MarkerArray marker_array_msg_;
     std::string name_;
 };
 }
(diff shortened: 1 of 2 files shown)


exotica/src/KinematicTree.cpp
```diff
@@ -311,11 +310,8 @@ void KinematicTree::resetModel()
     // Remove all CollisionShapes
     if (Server::isRos())
     {
-        visualization_msgs::MarkerArray msg;
-        visualization_msgs::Marker mrk;
-        mrk.action = 3;  // visualization_msgs::Marker::DELETEALL; // NB: enum only defined in ROS-J and newer, functionality still there
-        msg.markers.push_back(mrk);
-        shapes_pub_.publish(msg);
+        msg_marker.markers.emplace_back();
+        msg_marker.markers.end()->action = visualization_msgs::Marker::DELETEALL;
```
wxmerkt: Replace with `action = 3;  // visualization_msgs::Marker::DELETEALL; // NB: enum only defined in ROS-J and newer, functionality still there` to avoid breaking indigo support (it's the same).

exotica/include/exotica/KinematicTree.h
```diff
@@ -271,6 +272,7 @@ class KinematicTree : public Uncopyable
     std::vector<tf::StampedTransform> debugFrames;
     ros::Publisher shapes_pub_;
     bool debugSceneChanged;
+    visualization_msgs::MarkerArray msg_marker;
```
wxmerkt: `marker_array_msg_`

Issue #453, opened 2018-12-17
Not compiling LevenbergMarquardt throws _any_ solver on 16.04


wxmerkt:
The Levenberg-Marquardt solver is not compiled on 16.04 due to non-matching Eigen versions. However, as the solver configuration is still exported in the `package.xml`, _no_ solver can be loaded anymore as LevenbergMarquardt cannot be instantiated.

Potential solutions:
(a) Use Eigen3.3 as an external project.
(b) Create the export with CMake (empty or filled in).

PR #454, opened 2018-12-17
[exotica_levenberg_marquardt_solver] Use colPivHouseholderQr if Eigen < 3.3.0 (fix 16.04 bug)

wxmerkt:
As reported in #453: As we are not building the Levenberg-Marquardt solver on 16.04 _but_ maintain a `exotica_plugins.xml` file, `pyexotica` won't run any example. We have two solutions hereto (provide newer Eigen, or create the plugin-export file during configuration). This PR _may_ fix this issue but I do not have a way to test right now as I am on 16.04 over the holidays.

### DIFF
--- a/examples/exotica_examples/launch/PythonPlanLevenbergMarquardtIK.launch
+++ b/examples/exotica_examples/launch/PythonPlanLevenbergMarquardtIK.launch
@@ -1,0 +1,13 @@
+<launch>
+
+  <arg name="debug" default="false" />
+  <arg unless="$(arg debug)" name="launch_prefix" value="" />
+  <arg     if="$(arg debug)" name="launch_prefix" value="xterm -e gdb --args python" />
+
+  <param name="robot_description" textfile="$(find exotica_examples)/resources/robots/lwr_simplified.urdf" />
+  <param name="robot_description_semantic" textfile="$(find exotica_examples)/resources/robots/lwr_simplified.srdf" />
+
+  <node launch-prefix="$(arg launch_prefix)" pkg="exotica_examples" type="example_levenbergmarquardt.py" name="IKPlannerPythonExampleNode" output="screen" />
+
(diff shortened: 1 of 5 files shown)


wxmerkt:
This "fix" throws warnings though marked as error:
```
Skipped loading plugin with error: XML Document '/home/wxm/dev/exotica_ws/src/exotica/exotations/solvers/exotica_levenberg_marquardt_solver/exotica_plugins.xml' has no Root Element. This likely means the XML is malformed or missing..
```

VladimirIvan:
Can you either:
 A. Include only `CompleteOrthogonalDecomposition.h` from a newer Eigen? (copy/paste/include into src)
or
 B. Use a linear system solver base on a different decomposition (SVD, LU, ...)?

wxmerkt:
Option (A) does not work.

I have now added a compile-time macro to select the linear solver. On 16.04, we will use `colPivHouseholderQr` while opting for `completeOrthogonalDecomposition` if the Eigen version is recent enough. Please feel free to suggest [other linear solvers](https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html).

@christian-rauch Can you please have a look over? Thanks!

christian-rauch:
The main reason I selected `CompleteOrthogonalDecomposition` as solver was that for matrices with bad condition numbers, the alternative solvers (LDLT, LLT) produced unstable results (different results for the same input).

But from the comparison table in the Eigen documentation, it looks as if `CompleteOrthogonalDecomposition` and `ColPivHouseholderQR` have the same properties. So this would be fine by me.

The `example_levenbergmarquardt.xml` is using the `LevenbergMarquardtSolver` in its default configuration, which then behaves like the Gauss-Newton. You could add the damping parameter for demonstration purpose. Thanks for putting together the LM example @wxmerkt

PR #464, opened 2018-12-20
Load floating base limits from XML

VladimirIvan:
- Load floating base limits from XML (tag existed but it was unused)
- Updated examples
- Better solution for #444


VladimirIvan:
The initializer files already had these [in them](https://github.com/ipab-slmc/exotica/blob/master/exotica/init/SamplingProblem.in#L5) but the fields weren't used.

PR #465, opened 2018-12-20
Changed the underlying method used by the IK solver

### LINKED ISSUE #391: Implementation of PseudoInverse

### LINKED ISSUE #458: IK/AICO: Retire LAPACK
We rely on f2c'd version of LAPACK for the Cholesky factorisation in IK solver and AICO (coming with two dependencies, `libf2c2-dev` and `liblapack-dev`). This seems to sometimes cause issues w.r.t. performance and numerical stability (or now even throw segmentation faults on Chris' laptop). 

We should:

1. Benchmark the existing performance.
2. Replace with Eigen's symmetric-positive-definite or other linear solvers.
3. If comparable or better performance: replace.

VladimirIvan:
- Retired explicit pseudo-inverse computation, using a linear solver instead (solving for `(SJW)q=y`).
- Retired LAPACK, using `completeOrthogonalDecomposition` when available, `householderQr` otherwise.
- Runtime performance on both LWR and Valkyrie models is the same per single call.
- New solver is more numerically stable.
- Relates to #458 
- Fixes #391

Note:
- Setting C=0 means no regularisation. C=0.9999 (or similar) will regularise by virtually appending a scaled Identity matrix to the Jacobian. Values closer to 1 regularise more. The qNominal is used as a target to move towards for the added rows in the Jacobian. This can be achieved by adding an Identity task map to the problem though (only keeping this for convenience).
- `qNominal` is no longer minimised in the null space (see above).


exotations/solvers/exotica_ik_solver/include/exotica_ik_solver/ik_solver.h
```diff
@@ -59,15 +56,12 @@ class IKSolver : public MotionSolver, public Instantiable<IKSolverInitializer>
...
-    Eigen::MatrixXd PseudoInverse(Eigen::MatrixXdRefConst J);
     void ScaleToStepSize(Eigen::VectorXdRef xd);
 
     UnconstrainedEndPoseProblem_ptr prob_;  // Shared pointer to the planning problem.
 
-    Eigen::MatrixXd Cinv_;  //!< Weight Matrices
-    Eigen::MatrixXd C_;
-    Eigen::MatrixXd W_;
-    Eigen::MatrixXd Winv_;
+    Eigen::MatrixXd C_;  //!< \brief Regularisation (use values from interval <0, 1))
```
wxmerkt: This is wrong Doxygen. Should be `///<` (fixed the same issue yesterday in other places).

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -106,16 +72,33 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
 
     if (prob_->N != q0.rows()) throw_named("Wrong size q0 size=" << q0.rows() << ", required size=" << prob_->N);
 
-    const bool UseNullspace = prob_->qNominal.rows() == prob_->N;
+    Eigen::VectorXd qd, qNominal;
```
wxmerkt: `q_d`, `q_nominal`

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -106,16 +72,33 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
...
+    bool UseNullspace = false;
```
wxmerkt: `use_nullspace`

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -106,16 +72,33 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
...
+    }
+    else
+    {
+        qNominal = q0;
+    }
 
     solution.resize(1, prob_->N);
 
     Eigen::VectorXd q = q0;
     double error = INFINITY;
```
wxmerkt: `std::numeric_limits<double>::infinity()`

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -106,16 +72,33 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
...
+    double C = C_(0, 0);
+    Eigen::VectorXd yd;
+    Eigen::MatrixXd J;
+    if (C > 0)
+    {
+        yd = Eigen::VectorXd(prob_->JN + prob_->N);
+        J = Eigen::MatrixXd(prob_->JN + prob_->N, prob_->N);
+    }
     for (int i = 0; i < getNumberOfMaxIterations(); i++)
```
wxmerkt: `++i`

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -106,16 +72,33 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
...
+    else
+    {
+        qNominal = q0;
+    }
 
     solution.resize(1, prob_->N);
 
     Eigen::VectorXd q = q0;
     double error = INFINITY;
+    double C = C_(0, 0);
```
wxmerkt: `const double& C`

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
```diff
@@ -128,11 +111,24 @@ void IKSolver::Solve(Eigen::MatrixXd& solution)
...
+        else
+        {
+            yd = prob_->Cost.S * prob_->Cost.ydiff;
+            J = prob_->Cost.S * prob_->Cost.J * W_;
+        }
+
+#if EIGEN_VERSION_AT_LEAST(3, 3, 0)
+        qd = J.completeOrthogonalDecomposition().solve(yd);
+#else
+        qd = J.householderQr().solve(yd);
```
wxmerkt: Why not the `colPivHouseHolderQr` as used in Levenberg-Marquardt?

exotations/solvers/exotica_ik_solver/include/exotica_ik_solver/ik_solver.h
christian-rauch: This should be fine. I've used similar documentation strings before.
See: http://www.doxygen.nl/manual/docblocks.html (Putting documentation after members
, 3rd box).

christian-rauch:
Why are all these if/then switches for c required? What happends if C = Identity?
I.e. why can we not set matrix `C = I*c` and use it directly for `qd = (J = S * J * W*(1-C)).solve(yd)`?

examples/exotica_examples/scripts/example_attach.py
```diff
@@ -2,15 +2,16 @@
...
+import signal
 
 def handle_attaching(i, x, scene):
     # Update state
     scene.update(x)
     if i == 0:
         # Reset object pose
         scene.attach_object_local(
-            'Item', '', exo.KDLFrame([0.6, -0.3, 0.5, 0, 0, 0, 1]))
+            'Item', '', exo.KDLFrame([0.61, -0.3, 0.5, 0, 0, 0, 1]))
```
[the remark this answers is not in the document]
VladimirIvan: Well, I increased Rho=1e3 so now the solver reaches exactly where it's supposed to so the it touches the box on approach. I increased the distance otherwise OMPL start/goal state are invalid.

exotations/solvers/exotica_ik_solver/src/ik_solver.cpp
VladimirIvan: Sure, for consistency but `fullPivHouseHolderQr` and `householderQr` should be equally good for most problems.

VladimirIvan:
> Why are all these if/then switches for c required? What happends if C = Identity?
> I.e. why can we not set matrix `C = I*c` and use it directly for `qd = (J = S * J * W*(1-C)).solve(yd)`?

If `c>0` then
```
J= ⌈ SJW(1-c) ⌉, yd= ⌈ Sy_diff     ⌉
   ⌊ Ic       ⌋      ⌊ q-q_nominal ⌋ 
```
This is a hack to get the solution to scale down `SJW` by `1-c` while scaling up the the bottom part of `J`. This way `c` will blend between minimising the cost (`y_diff`) and minimising the distance to nominal pose. This will also add non-zero elements into `J` so singular problems will be biased towards `q_nominal`. The regularised pseudo-inverse handled this better (without adding `Ic` into `J`) but it was also less stable because of explicit computation of `Jinv`. If there is a better way to achieve this regularisation I'm happy to use that.

If `c<=0`, `C` is completely dropped from the equation for performance.

What you propose (without the if-s) would only scale down the J, it wouldn't add non-zero elements into it. I keep that scaling (`SJW(1-c)`) so that you have some notion what value of `c` to choose (e.g., between 0 and 1). It's still not perfect because when Rho (and S) are high, `c` also has to be scaled to achieve the same regularisation (e.g., Rho=1 >> c= 0.9, Rho=1e-3, c=0.999). This wasn't the case with the pseudo-inverse.

wxmerkt:
Actually, while we are updating it: @VladimirIvan could you please add some doxygen to the header? About the implementation, with references, etc.

wxmerkt:
This is a very positive change performance-wise:

On my Valkyrie whole-body IK benchmark:

- Before this change: ~200ms
- After this change: ~40ms

Issue #456, opened 2018-12-18
PRM::grow_roadmap and others broken


wxmerkt:
This is directly related to #445 and #444:

- Because the solver isn't setup before solve, calling `grow_roadmap` results in a segmentation fault.
- This breaks `example_prm`
- The actual cause is that `setup()` hasn't been called. This can be checked with `isSetup()`.
- Also related to #435 - likely broken there.

Backtrace:
```
[New Thread 0x7fffb3fff700 (LWP 17198)]
[ INFO] [1545135698.834008741]: Loading robot model 'lwr'...

Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffcbd706a0 in ompl::base::Planner::isSetup() const () from /opt/ros/kinetic/lib/x86_64-linux-gnu/libompl.so.12
(gdb) backtrace
#0  0x00007fffcbd706a0 in ompl::base::Planner::isSetup() const () from /opt/ros/kinetic/lib/x86_64-linux-gnu/libompl.so.12
#1  0x00007fffcbd9a063 in ompl::geometric::PRM::growRoadmap(ompl::base::PlannerTerminationCondition const&) () from /opt/ros/kinetic/lib/x86_64-linux-gnu/libompl.so.12
#2  0x00007fffcbd9a16c in ompl::geometric::PRM::growRoadmap(double) () from /opt/ros/kinetic/lib/x86_64-linux-gnu/libompl.so.12
#3  0x00007fffcc2f2fa2 in exotica::PRM::GrowRoadmap (this=<optimised out>, t=<optimised out>)```

wxmerkt:
@VladimirIvan @christian-rauch The source of this bug is that the state space and simple setup aren't finalised ("setup") until the first call of `Solve`. Thus, growing a roadmap before the first solve has been called (as in `example_prm.py`) causes a segmentation fault. There are a few issues that prevent easy fixes:

(a) We moved to this in order to be able to set correct bounds for the problem. That was broken.
(b) Previously when we first created problems and solvers specifically, we could change the problem before calling `specifyProblem` to the solver - which circumvented this issue. Now the standard way is to use `loadSolver` which specifies the problem automatically.
(c) Calling setup manually after checking isSetup is also crashing :'(
(d) We cannot reset the state space without resetting the simple setup - which is resetting everything.

VladimirIvan:
> (d) We cannot reset the state space without resetting the simple setup - which is resetting everything.

[...]

PR #470, opened 2018-12-21
[exotica_examples] Renaming and code standardization

### LINKED ISSUE #463: Dependency on Joy package
`PythonPlanIKEffAxisAlignment.launch`, `PythonPlanIKPointToLine.launch`, `PythonPlanIKWithJntSmooth.launch` depend on a physical joystick being plugged in and it has to be a very specific type of joystick.
These examples are therefore impossible to maintain without the hardware. 
All of these examples can either use generated input or use keyboard.

VladimirIvan:
- Moved examples from `examples/exotica_examples` to `exotica_examples`.
- Renamed all files consistently (including tests).
- Fixed broken examples (including workaround for #456).
- Removed dependency on joy and replaced it with an interactive marker (fixes #463).
- Visually checked all examples.
- Added python module with `TargetMarker` helper class for setting up interactive markers (used in some IK examples now).
- Updated the `run_examples` script to run all existing examples.

Notes:
 - `python_prm` runs very slowly.


exotica_examples/CMakeLists.txt
```diff
@@ -51,13 +44,13 @@ install(DIRECTORY resources/ DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/res
 FILE(GLOB python_examples "${CMAKE_CURRENT_SOURCE_DIR}/scripts/*")
 catkin_install_python(PROGRAMS ${python_examples} DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
+catkin_python_setup()
```
wxmerkt: I always thought this has to come _before_ `catkin_package()` but now I can't find this in the documentation anymore (cf. [here](http://docs.ros.org/melodic/api/catkin/html/user_guide/setup_dot_py.html))

exotica_examples/resources/rviz.rviz
```diff
@@ -186,6 +188,6 @@ Window Geometry:
     collapsed: false
   Views:
     collapsed: false
-  Width: 1855
-  X: 1025
-  Y: 536
+  Width: 3738
```
wxmerkt: You now need a 4K screen for our examples, duh.

exotica_examples/scripts/example_prm
```diff
@@ -12,6 +12,11 @@
 # You have to call prm.clear_query() between runs.
 # Call prm.clear() to discard the roadmap.
 
+# Temporary fix: call solve() to setup the OMPL problem,
```
wxmerkt: Add a reference to the issue

exotica_examples/src/exotica_examples_py/target_marker.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+from interactive_markers.interactive_marker_server import *
+from visualization_msgs.msg import *
+from geometry_msgs.msg import Pose, Point, Quaternion
+import PyKDL as kdl
+import pyexotica as exo
+
```
wxmerkt: Add an `__all__ = [list_to_pose, TargetMarker]` here - so when people do `from exotica_examples_py import *` they do not get transient access to PyKDL etc.

exotica_examples/package.xml
```diff
@@ -20,10 +20,11 @@
...
   <exec_depend>exotica_time_indexed_rrt_connect_solver</exec_depend>
-  <exec_depend>joy</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>rviz</exec_depend>
-  <exec_depend>joy</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
+  <exec_depend>visualization_msgs</exec_depend>
+  <exec_depend>interactive_markers</exec_depend>
   <test_depend>rostest</test_depend>
   <test_depend>exotica_val_description</test_depend>
```
wxmerkt: PyKDL dependency is missing

wxmerkt:
The look_at example doesnt work for me though hmmm it's only moving the first two joints

exotica_examples/CMakeLists.txt
VladimirIvan: The docs say `When using the genmsg macro, ordering constraints exist`
and they provide this order:
```
project(...)
...
find_package(catkin ...)
...
catkin_python_setup()
...
generate_messages()
...
catkin_package()
```

The auto-generated CMakeLists also has it in that place.

VladimirIvan:
> The look_at example doesnt work for me though hmmm it's only moving the first two joints

That's how it was designed to work. I don't like the look_at and the point_to_line examples for the same reason. They don't show what the task map does in an intuitive way. @cmower

exotica_examples/CMakeLists.txt
wxmerkt: Thanks, I knew I read it somewhere but couldnt find the reference!

wxmerkt:
Hmmm, then the look_at name is a bit misleading (and not what a gaze, the alternative name suggestion, would do).

PR #471, opened 2018-12-22
Add default colour for addObjectToEnvironment

wxmerkt:
`Eigen::Vector4d()` is pointing to uninitialised memory. We should use a default colour, and here we use the same as as in the `KinematicTree` (a grey; noticed when generating Python documentation)


christian-rauch:
Isn't `Eigen::Vector4d()` constructing a `[0,0,0,0]` vector?
By default, if no colour was provided, the markers were published with the gray colour (50% black). But making this more explicit if definitely better.

wxmerkt:
No, unless you select `Eigen::Vector4d::Zero()`, `Eigen::Vector4d()` will only create the data struct pointing to unitialised memory (something e-300). This is one of the biggest Eigen gotchas ;-)

PR #474, opened 2018-12-28
[exotica] Renamed to exotica_core, standardised file names

VladimirIvan:
- Renamed `exotica` to `exotica_core`
- Renamed files according to the the style guide


exotica_core/include/exotica_core/problems/bounded_end_pose_problem.h
```diff
@@ -33,10 +33,10 @@
 #ifndef BOUNDEDENDPOSEPROBLEM_H_
```
wxmerkt: None of the files have the header guards fixed. In other packages we now adopted `EXOTICA_PACKAGE_NAME_HEADER_NAME_H_`, e.g., `EXOTICA_AICO_SOLVER_MATH_OPERATIONS_H_`

exotica_core/include/exotica_core/problems/time_indexed_sampling_problem.h
```diff
@@ -33,10 +33,10 @@
 #ifndef INCLUDE_EXOTICA_PROBLEMS_TIMEINDEXEDSAMPLINGPROBLEM_H_
```
wxmerkt: E.g. this one - the header guard is absolutely none-consistent

exotica_core/src/problems/end_pose_problem.cpp
```diff
@@ -30,9 +30,9 @@
  *
  */
 
-#include <exotica/Problems/EndPoseProblem.h>
-#include <exotica/Setup.h>
-#include <exotica/TaskInitializer.h>
+#include <exotica_core/TaskInitializer.h>
```
wxmerkt: I always split off groups of includes by an empty line:
1. System headers (C first, then C++)
2. exotica_core includes
3. Package or other dependency includes
4. Initializers

exotica_core/src/scene.cpp
```diff
@@ -262,7 +262,7 @@ void Scene::publishProxies(const std::vector<CollisionProxy>& proxies)
 visualization_msgs::Marker Scene::proxyToMarker(const std::vector<CollisionProxy>& proxies, const std::string& frame)
 {
     visualization_msgs::Marker ret;
-    ret.header.frame_id = "exotica/" + frame;
+    ret.header.frame_id = "exotica_core/" + frame;
```
wxmerkt: I don't think we want to change this one.

exotica_core/src/setup.cpp
```diff
@@ -107,7 +107,7 @@ std::vector<std::string> Setup::getSolvers() { return Instance()->solvers_.getDe
 std::vector<std::string> Setup::getProblems() { return Instance()->problems_.getDeclaredClasses(); }
 std::vector<std::string> Setup::getMaps() { return Instance()->maps_.getDeclaredClasses(); }
 std::vector<std::string> Setup::getCollisionScenes() { return Instance()->scenes_.getDeclaredClasses(); }
-Setup::Setup() : solvers_("exotica", "exotica::MotionSolver"), maps_("exotica", "exotica::TaskMap"), problems_(PlanningProblem_fac::Instance()), scenes_("exotica", "exotica::CollisionScene")
+Setup::Setup() : solvers_("exotica_core", "exotica::MotionSolver"), maps_("exotica_core", "exotica::TaskMap"), problems_(PlanningProblem_fac::Instance()), scenes_("exotica_core", "exotica::CollisionScene")
```
wxmerkt: I don't think we want to change this one - I'd prefer the namespace for plugins to remain `exotica`.

exotica_core/src/tools.cpp
```diff
@@ -30,7 +30,7 @@
  *
  */
 
-#include "exotica/Tools.h"
+#include "exotica_core/tools.h"
 #include <cxxabi.h>  //!< The demangler for gcc... this makes this system dependent!
```
wxmerkt: Change the `//!<` to a normal `// NB:` - otherwise it messes our doxygen up

exotations/exotica_collision_scene_fcl/package.xml
```diff
@@ -9,11 +9,12 @@
...
   <buildtool_depend>catkin</buildtool_depend>
-  <depend>exotica</depend>
+  <depend>exotica_core</depend>
   <depend>geometric_shapes</depend>
   <depend>libfcl-dev</depend>
+  <depend>roscpp</depend>
 
   <export>
-    <exotica plugin="${prefix}/exotica_plugins.xml" />
+    <exotica_core plugin="${prefix}/exotica_plugins.xml" />
```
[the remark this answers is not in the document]
VladimirIvan: The tag name has to match the package name which it extends. That is how they match the plugins to the the library/node they extend.

exotica_core/src/scene.cpp
VladimirIvan: Good spot

wxmerkt:
Merging. Note the header guards aren’t as desired (not including package names)

PR #482, opened 2019-01-08
throw if unsupported rotation type has been selected

christian-rauch:
Throw an exception if an unsupported rotation type is selected (e.g. for catching typos like "rpy" or "Quaternions") instead of using RPY by default.

### DIFF
--- a/exotations/exotica_core_task_maps/src/eff_orientation.cpp
+++ b/exotations/exotica_core_task_maps/src/eff_orientation.cpp
@@ -42,6 +42,10 @@ void EffOrientation::Instantiate(EffOrientationInitializer& init)
     {
         rotation_type_ = RotationType::QUATERNION;
     }
+    else if (init.Type == "RPY")
+    {
+        rotation_type_ = RotationType::RPY;
+    }
     else if (init.Type == "ZYX")
     {
         rotation_type_ = RotationType::ZYX;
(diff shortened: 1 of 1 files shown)


wxmerkt:
Thank you - much needed improvement :-)

christian-rauch:
I also recognised that the task map is not checking the size of its input. E.g. you can provide a 4D quaternion for a 3D RPY.

wxmerkt:
Good call, could you please add those checks? Thank you!

christian-rauch:
I was confused.
The conversion has nothing to do with the rotation type parameter of the `EffOrientation` task map. The task map already gets the KDL frame, respectively the rotation matrix. The conversion is somewhere done in `exotica/Frame`.

wxmerkt:
Have a look at [`GetFrameAsRotation`](http://ipab-slmc.github.io/exotica/doxygen_cpp/namespaceexotica.html#aae93646f46418711fd0f9a3bf61cca17) and [`GetFrameAsVector`](http://ipab-slmc.github.io/exotica/doxygen_cpp/namespaceexotica.html#a2d49c925955346266f79f8b47cfb5e52):

https://github.com/ipab-slmc/exotica/blob/cad9fbad9906cefed16cf0d16fbe05dc9c163a23/exotica_core/src/tools/conversions.cpp#L100-L113

I don't think we want the proposed solution in #484 - it suffers from the gimbal lock and takes away the advantages of supporting multiple rotation types.

christian-rauch:
The automatic conversion I mentioned is done in `GetFrame`:
https://github.com/ipab-slmc/exotica/blob/cad9fbad9906cefed16cf0d16fbe05dc9c163a23/exotica_core/src/tools/conversions.cpp#L51-L68

PR #484, opened 2019-01-08
use RPY task space vector for all rotation types

### LINKED ISSUE #483: Use of rotation costs in EffOrientation
The `EffOrientation` task map converts any provided rotation format (Quaternion, Euler angles, ...) into a variable length task space vector (e.g Quaternion -> 4D, RPY -> 3D, Rotation matrix -> 9D, axis-angle -> 3D) but uses the rotation part of the twist (3D) of the Jacobian as gradient for all of these task vectors.

This doesn't seem appropriate. In fact, this is only working if the RPY rotation type is used since the rotation of the twist is represented as RPY.
If the rotation part of the twist is used as gradient, the task vector needs to be constructed from the corresponding 3 Euler angles.

As alternative, the magnitude of a rotation is usually the absolute angle of the axis-angle representation. However, the gradients for this cannot directly be obtained from the Jacobian.

christian-rauch:
Convert all orientations to RPY and use this as 3D task space vector. Fixes #483 .

This would also invalidate PR https://github.com/ipab-slmc/exotica/pull/482.

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/eff_orientation.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/eff_orientation.h
@@ -50,10 +50,6 @@ public:
     int TaskSpaceJacobianDim() override;
 
     std::vector<TaskVectorEntry> GetLieGroupIndices() override;
-    RotationType rotation_type_ = RotationType::RPY;  // TODO: Make private with getter
-
-private:
-    int stride_;
 };
 }
 
--- a/exotations/exotica_core_task_maps/init/eff_orientation.in
+++ b/exotations/exotica_core_task_maps/init/eff_orientation.in
@@ -1,5 +1,3 @@
 class EffOrientation
 
 extend <exotica_core/task_map>
-
-Optional std::string Type = "RPY";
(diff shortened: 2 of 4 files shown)


VladimirIvan:
After the discussion in #483, why remove the option to provide goals as RPY (or other)?
Furthermore, for some reason, the quaternion has lower accuracy than RPY (we had to reduce the eps in the tests).
This PR is removing a feature without a reason and it makes EffOrientation and EffFrame inconsistent.

christian-rauch:
Yes. I wasn't aware that the task space vector can be manipulated manually (via `SetGoal`). Of course, if you want the user to have raw access to the task space vector, he needs to know the representation of the orientation, respectively allow to define the representation.

This PR was created with having in mind, that there is now raw access to the task space vector and that the serialisation of the rotation is just an (arbitrary) intermediate representation.

christian-rauch:
I am closing this until we found a solution to automatically determine the rotation type in `PlanningProblem::SetGoal` (see: https://github.com/ipab-slmc/exotica/issues/483#issuecomment-453073678) and use a fixed orientation representation in the task space vector.

PR #487, opened 2019-01-09
seed for sampling joint states in KinematicTree

christian-rauch:
Method `SetSeed` for setting seed of random number generator in `KinematicTree`.

### DIFF
--- a/exotica_core/include/exotica_core/kinematic_tree.h
+++ b/exotica_core/include/exotica_core/kinematic_tree.h
@@ -210,6 +210,7 @@ public:
 
     std::map<std::string, shapes::ShapeType> GetCollisionObjectTypes();
 
+    void SetSeed(const uint_fast32_t seed) { generator_.seed(seed); }
     /// Random state generation
     Eigen::VectorXd GetRandomControlledState();
 
--- a/exotica_core/include/exotica_core/kinematic_tree.h
+++ b/exotica_core/include/exotica_core/kinematic_tree.h
@@ -210,6 +210,8 @@ public:
 
     std::map<std::string, shapes::ShapeType> GetCollisionObjectTypes();
 
+    /// @brief SetSeed sets the seed of the random generator for deterministic joint state sampling
+    /// @param seed unsigned integer
     void SetSeed(const uint_fast32_t seed) { generator_.seed(seed); }
     /// Random state generation
     Eigen::VectorXd GetRandomControlledState();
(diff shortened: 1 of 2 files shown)


christian-rauch:
The `std::mt19937` does not store the seed it was initialised with. Or at least there is no public method/member to access this.
We would need to store the seed as a member of `KinematicTree` and this would also not be defined if `SetSeed` wouldn't have been called before.

christian-rauch:
I didn't know that there are different doxygen styles per language.
QtCreator is actually using this style per default. E.g. you type `/**` + Enter and it creates the template automatically. In this case:
```
/**
 * @brief SetSeed
 * @param seed
 */
```
I very like this feature because automatically includes all parameters and the return value of a method.

Do you insist on the triple slash (`///`) style? :-)

wxmerkt:
I like the same auto-completion feature and would be happy with either - Vlad does insist ;)

christian-rauch:
Oh well, here you go :-)

exotica_core/include/exotica_core/kinematic_tree.h
```diff
@@ -210,6 +210,9 @@ class KinematicTree : public Uncopyable
 
     std::map<std::string, shapes::ShapeType> GetCollisionObjectTypes();
 
+     /// @brief SetSeed set the seed of the random generator for deterministic joint state sampling
```
wxmerkt: "sets"

exotica_core/include/exotica_core/kinematic_tree.h
```diff
@@ -210,6 +210,9 @@ class KinematicTree : public Uncopyable
...
+     /// @param seed unsigned integer
+    void SetSeed(const uint_fast32_t seed) { generator_.seed(seed); }
```
wxmerkt: `const uint_fast32_t&` (pass by reference, minor here but let's be consistent)

christian-rauch: What's the advantage of using a reference here? `uint32` and its reference have the same size, i.e. there is no performance benefit.

wxmerkt: You're right and I missed something here. We shouldn't do it for fundamental types (cf. [here](http://www.learncpp.com/cpp-tutorial/73-passing-arguments-by-reference/)) and apart from double and long long the types are smaller than their reference (cf. [here](https://softwareengineering.stackexchange.com/a/356200)). Also apparently it may break compiler optimisation by aliasing (duh). Sorry I didn't realise that earlier, let's go back to how you initially had it and apologies for the extra trouble :'(.

christian-rauch: No problem :-)

christian-rauch:
This should be fine now.

wxmerkt:
Thank you very much :-)

PR #495, opened 2019-01-17
Finite differencing bug fix

cmower:
* was not updating the kinematic tree
* implementation of backward finite differencing was incorrect
* new implementation has been validated using `example_ik`, I forced `EffFrame` task map to use finite differencing rather than analytical jacobian and all worked as expected (`EffFrame` was reset before commiting changes to exotica_core)
* code formatting

# Checklist

### DIFF
--- a/exotica_core/src/task_map.cpp
+++ b/exotica_core/src/task_map.cpp
@@ -73,24 +73,33 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
     if (jacobian.rows() != TaskSpaceDim() && jacobian.cols() != x.rows())
         ThrowNamed("Jacobian dimension mismatch!");
 
-    // Store original x (needs to be reset later).
-    Eigen::VectorXd x_original(x);
-    Update(x_original, Phi);
-
-    // Backward finite differencing.
+    // Set constants
     constexpr double h = 1e-6;
-    Eigen::VectorXd x_tmp;
-    Eigen::VectorXd phi_original(Phi);
-    for (int i = 0; i < TaskSpaceDim(); ++i)
+    constexpr double h_inverse = 1.0 / h;
+
+    // Compute x/Phi using forward mapping (no jacobian)
+    Update(x, Phi);
+
+    // Setup for gradient estimate
+    Eigen::VectorXd x_backward(x.size()), Phi_backward(TaskSpaceDim());
+
+    // Backward finite differencing
+    for (int i = 0; i < jacobian.cols(); ++i)
     {
-        x_tmp = x;
-        x_tmp(i) -= h;
-        Update(x_tmp, Phi);
-        jacobian.row(i) = (1 / h) * (phi_original - Phi);
+        // Compute and set x_backward as model state
+        x_backward = x;
+        x_backward(i) -= h;
+        scene_->GetKinematicTree().Update(x_backward);
+
+        // Compute phi_backward using forward mapping (no jacobian)
+        Update(x_backward, Phi_backward);
+
+        // Compute gradient estimate
+        jacobian.col(i) = h_inverse * (Phi - Phi_backward);
     }
 
-    // Finally, reset with original value again.
-    Update(x_original, Phi);
+    // Reset model state
+    scene_->GetKinematicTree().Update(x);
 }
 
 void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::MatrixXdRef jacobian, HessianRef hessian)


exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
-        Update(x_tmp, Phi);
-        jacobian.row(i) = (1 / h) * (phi_original - Phi);
+        // Setup for gradient estimation
+        Eigen::VectorXd x_down = x;
+        x_down(i) -= h;
+        Eigen::VectorXd phi_down;
+        phi_down.resize(ntask);
+
+        // Set x_down as model state
+        scene_->GetKinematicTree().SetModelState(x_down);
```
wxmerkt: `SetModelState` or `Update`? I believe this should be `Update`, otherwise will break floating-base or versions where only a subset of joints is controlled.

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
+
+    // Backward finite differencing
+    for (int i = 0; i < ndof; ++i)
     {
-        x_tmp = x;
-        x_tmp(i) -= h;
-        Update(x_tmp, Phi);
-        jacobian.row(i) = (1 / h) * (phi_original - Phi);
+        // Setup for gradient estimation
+        Eigen::VectorXd x_down = x;
```
wxmerkt: Please use a different suffix than `down`.

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
-
-    // Backward finite differencing.
+    // Set constants
     constexpr double h = 1e-6;
-    Eigen::VectorXd x_tmp;
-    Eigen::VectorXd phi_original(Phi);
-    for (int i = 0; i < TaskSpaceDim(); ++i)
+    constexpr double hi = 1.0 / h;
+    int ntask = TaskSpaceDim();
+    int ndof = jacobian.cols();
```
wxmerkt: `ndof` is only used once, no need to maintain a variable for that. Directly use it in the `for`

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
-    Update(x_original, Phi);
-
-    // Backward finite differencing.
+    // Set constants
     constexpr double h = 1e-6;
-    Eigen::VectorXd x_tmp;
-    Eigen::VectorXd phi_original(Phi);
-    for (int i = 0; i < TaskSpaceDim(); ++i)
+    constexpr double hi = 1.0 / h;
+    int ntask = TaskSpaceDim();
```
wxmerkt: Only used once, use it directly below where needed

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
-    Eigen::VectorXd x_original(x);
-    Update(x_original, Phi);
-
-    // Backward finite differencing.
+    // Set constants
     constexpr double h = 1e-6;
-    Eigen::VectorXd x_tmp;
-    Eigen::VectorXd phi_original(Phi);
-    for (int i = 0; i < TaskSpaceDim(); ++i)
+    constexpr double hi = 1.0 / h;
```
wxmerkt: `h_inv` or `h_inverse`

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,36 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
+        phi_down.resize(ntask);
+
+        // Set x_down as model state
+        scene_->GetKinematicTree().SetModelState(x_down);
+
+        // Compute phi_down using forward mapping (no jacobian)
+        Update(x_down, phi_down);
+
+        // Compute gradient estimate
+        jacobian.col(i) = hi * (Phi - phi_down);
```
wxmerkt: Capitalisation of `Phi` should be consistent.

exotica_core/src/task_map.cpp
cmower: * I have tested both `scene_->GetKinematicTree().Update(x)` and `scene_->GetKinematicTree().SetModelState(x)` on kuka and valkyrie models
* Both `SetModelState` and `Update` work for the simplified kuka but `SetModelState` fail on valkyrie, throws an error re wrong start state size
* Will update to use `scene_->GetKinematicTree().Update(x)`

wxmerkt: Yep, that is expected... because SetModelState requires the full state to be set. While x is only the currently active planning group So for wherever planning group != whole-body state (e.g. only planning for a subgroup) this would fail.

exotica_core/src/task_map.cpp
```diff
@@ -73,24 +73,34 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
...
+    for (int i = 0; i < jacobian.cols(); ++i)
     {
-        x_tmp = x;
-        x_tmp(i) -= h;
-        Update(x_tmp, Phi);
-        jacobian.row(i) = (1 / h) * (phi_original - Phi);
+        // Setup for gradient estimation
+        Eigen::VectorXd x_backward = x;
+        x_backward(i) -= h;
+        Eigen::VectorXd Phi_backward;
```
wxmerkt: Don't do this in two operations and don't reallocate at every iteration. Define Phi_backward(TaskSpaceDim) outside the for-loop (same for x_backward)

exotica_core/src/task_map.cpp
```diff
@@ -77,19 +77,19 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
     constexpr double h = 1e-6;
     constexpr double h_inverse = 1.0 / h;
 
-    // Compute x/Phi using forward mapping (no jacobain)
+    // Compute x/Phi using forward mapping (no jacobian)
     Update(x, Phi);
 
+    // Setup for gradient estimate
+    Eigen::VectorXd x_backward, Phi_backward;
+    Phi_backward.resize(TaskSpaceDim());
```
wxmerkt: You can do this in one line: `Eigen::VectorXd Phi_backward(TaskSpaceDim());`

wxmerkt: Similarly `x_backward(x.size())`.

Issue #437, opened 2018-12-12
Cast dynamically loaded plugins to derived class to access non-virtual functions


wxmerkt:
Noticed while writing a unit test for a task map:

- We load taskmaps (and many other things) as plugins where most of the relevant functions are provided via polymorphism (virtual in the interface, override in the derived class). 
- However, some taskmaps require explicit additional methods that are not part of the interface but only in the derived class (`getPoint`, `setPoint`, `SetPreviousJointState`).
- This works fine in the Python module as it is compiled with all the source files.
- In C++, when we load (via `pluginlib` -> `class_loader` -> `libpoco-dev`), we cannot cast the pointer of a taskmap (`Taskmap_ptr`) to its derived type to access the derived-only methods.

PR #496, opened 2019-01-17
Adding manipulability task map

cmower:
* adds the task map for manipulability using Yoshikawa's formula
* currently requires finite differencing
* added an example to accompany task map

# Checklist

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
@@ -49,7 +49,7 @@ namespace exotica
 ///
 /// Note that
 ///   - the assocaited value for \f$\rho\f$ <b>must</b> be negative in order to maximize the manipulability, and
-///   - derivatives of \f$\Phi\f$ are computed using finite differences. 
+///   - derivatives of \f$\Phi\f$ are computed using finite differences.
 class Manipulability : public TaskMap, public Instantiable<ManipulabilityInitializer>
 {
 public:
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
@@ -1,0 +1,68 @@
+//
+// Copyright (c) 2018, University of Edinburgh
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//  * Redistributions of source code must retain the above copyright notice,
+//    this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of  nor the names of its contributors may be used to
+//    endorse or promote products derived from this software without specific
+//    prior written permission.
+//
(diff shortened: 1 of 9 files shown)


exotations/exotica_core_task_maps/src/manipulability.cpp
```diff
@@ -0,0 +1,59 @@
...
+void Manipulability::Instantiate(ManipulabilityInitializer& init)
+{
+    n_end_effs_ = frames_.size();
+}
+
+int Manipulability::TaskSpaceDim()
+{
+    return n_end_effs_;
+}
+}
```
wxmerkt: `}  // namespace exotica`

exotations/exotica_core_task_maps/src/manipulability.cpp
```diff
@@ -0,0 +1,59 @@
...
+Manipulability::Manipulability() = default;
+Manipulability::~Manipulability() = default;
+
+void Manipulability::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi)
+{
+    if (phi.rows() != TaskSpaceDim()) ThrowNamed("Wrong size of phi!");
+
+    for (int i = 0; i < n_end_effs_; ++i)
+    {
+        Eigen::MatrixXd J = kinematics[0].jacobian[i].data;
```
wxmerkt: `const Eigen::MatrixXd&`

cmower:
Todo
- [x] address inline comments
- [ ] add unit test

exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
```diff
@@ -0,0 +1,70 @@
...
+{
+public:
+    Manipulability();
+    virtual ~Manipulability();
+
+    void Instantiate(ManipulabilityInitializer& init) override;
+    void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi) override;
+    int TaskSpaceDim() override;
+
+    double get_manipulability_index(int& i);
```
wxmerkt: Don't use reference for basic types. Use `const` though, i.e., `const int i`.

exotations/exotica_core_task_maps/src/manipulability.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return std::sqrt((J * J.transpose()).determinant());
+}
+
+void Manipulability::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi)
+{
+    if (phi.rows() != TaskSpaceDim()) ThrowNamed("Wrong size of phi!");
+
+    for (int i = 0; i < n_end_effs_; ++i)
+    {
+        phi(i) = get_manipulability_index(i);
```
wxmerkt: Do you need the exposed-to-Python? You can get the `Phi` of the taskmap... and I'd rather inline the function above than have this call here.

cmower: I thought it would be useful for users to have the facility to get the manipulability index in python, perhaps for logging or further analysis purposes.

wxmerkt: They can via the `Phi`: you can get the start and length of the task map and thus extract it there. It _may_ be but it's limited to Python only (these calls do not work in C++, cf. #437) and I'd like the declaration of the function to be `inline` if we keep it here.

cmower: Ok, will revert changes

wxmerkt:
Thank you! #495 is now merged. Can you please re-base and squash this branch on top of master?

PR #503, opened 2019-01-29
Adds DynamicsSolver plug-in class and DynamicTimeIndexedShootingProblem

wxmerkt:
This is the first step to adding problems involving dynamics into Exotica. As part of this, this pull request:

1. Introduces a new plug-in class, `DynamicsSolver`, based off the abstract `AbstractDynamicsSolver` class. The dynamics can hereby be either from an implementation of the Articulated Body Algorithm (forward dynamics), or explicit hard-coded dynamics. These are to be implemented in plug-ins. The dynamics solver contains different integration scheme - for now only Euler and explicit trapezoidal (more to be added - RK4, RK45).
2. As a first type of such a plug-in, the canonical double integrator (i.e., `DoubleIntegratorDynamicsSolver`)
3. A first trajectory optimisation problem containing dynamics (indirect optimal control - shooting problem): `DynamicTimeIndexedShootingProblem`.
4. A simple example to run a double integrator dynamics solver on the LWR using Python, and solving an indirect optimal control problem using finite differences (with a scipy solver - the first example for showing the use of an external/Python solver in combination with an Exotica planning problem).

This PR leaves lots of todo items and good starting points.

### DIFF
--- a/exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
+++ b/exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
@@ -74,7 +74,7 @@ void DynamicTimeIndexedShootingProblem::Instantiate(DynamicTimeIndexedShootingPr
         }
         else
         {
-            ThrowNamed("Q dimension mismatch! Expected " << dynamics_solver_->get_num_controls() << ", got " << init_.R.rows());
+            ThrowNamed("R dimension mismatch! Expected " << dynamics_solver_->get_num_controls() << ", got " << init_.R.rows());
         }
     }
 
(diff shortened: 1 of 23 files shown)


exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
```diff
@@ -0,0 +1,295 @@
...
+    R_ = init_.R_rate * Eigen::MatrixXd::Identity(dynamics_solver_->get_num_controls(), dynamics_solver_->get_num_controls());
+    if (init_.R.rows() > 0)
+    {
+        if (init_.R.rows() == dynamics_solver_->get_num_controls())
+        {
+            R_.diagonal() = init_.R;
+        }
+        else
+        {
+            ThrowNamed("Q dimension mismatch! Expected " << dynamics_solver_->get_num_controls() << ", got " << init_.R.rows());
```
VladimirIvan: `R` instead of `Q`

PR #506, opened 2019-01-30
Create scene method

christian-rauch:
This adds the `CreateScene` method to instantiate a scene from an Initializer.

It can be used, e.g. for getting the model links before a problem is created:
```Python
model_links = exo.Setup.create_scene(('exotica/Scene', scene)).get_model_link_names()
```

### DIFF
--- a/exotica_core/include/exotica_core/setup.h
+++ b/exotica_core/include/exotica_core/setup.h
@@ -78,12 +78,6 @@ public:
         ret->InstantiateInternal(init);
         return ret;
     }
-    static std::shared_ptr<exotica::TaskMap> CreateMap(const Initializer& init)
-    {
-        std::shared_ptr<exotica::TaskMap> ret = ToStdPtr(Instance()->maps_.createInstance(init.GetName()));
-        ret->InstantiateInternal(init);
-        return ret;
-    }
     static std::shared_ptr<exotica::PlanningProblem> CreateProblem(const Initializer& init)
     {
         std::shared_ptr<exotica::PlanningProblem> ret = Instance()->problems_.CreateInstance(init.GetName());
@@ -108,11 +102,19 @@ public:
     }
 
 private:
+    friend PlanningProblem;
     Setup();
     static std::shared_ptr<Setup> singleton_initialiser_;
     // Make sure the singleton does not get copied
     Setup(Setup const&) = delete;
     void operator=(Setup const&) = delete;
+
+    static std::shared_ptr<exotica::TaskMap> CreateMap(const Initializer& init)
+    {
+        std::shared_ptr<exotica::TaskMap> ret = ToStdPtr(Instance()->maps_.createInstance(init.GetName()));
+        ret->InstantiateInternal(init);
+        return ret;
+    }
 
     pluginlib::ClassLoader<exotica::MotionSolver> solvers_;
     pluginlib::ClassLoader<exotica::TaskMap> maps_;
--- a/exotica_core/include/exotica_core/setup.h
+++ b/exotica_core/include/exotica_core/setup.h
@@ -91,6 +91,22 @@ public:
         return ret;
     }
 
+    ///
+    /// \brief CreateScene instantiate a scene from an initialiser
+    ///     The returned scene is independent of the internal EXOTica solver
+    ///     or problem state. It can only be used to access the parsed information
+    ///     like joint and link names or the kinematics. Changes to the scene
+    ///     will not affect the solver or problem.
+    /// \param init scene initialiser
+    /// \return a shared pointer to the scene
+    ///
+    static exotica::ScenePtr CreateScene(const Initializer& init)
+    {
+        exotica::ScenePtr ret = std::make_shared<exotica::Scene>();
+        ret->InstantiateInternal(init);
+        return ret;
+    }
+
 private:
     Setup();
     static std::shared_ptr<Setup> singleton_initialiser_;
--- a/exotica_python/src/pyexotica.cpp
+++ b/exotica_python/src/pyexotica.cpp
@@ -497,7 +497,6 @@ PYBIND11_MODULE(_pyexotica, module)
     setup.def_static("create_solver", [](const Initializer& init) { return Setup::CreateSolver(init); }, py::return_value_policy::take_ownership);    // "Creates an instance of the solver identified by name parameter.", py::arg("solverType"), py::arg("prependExoticaNamespace"));
     setup.def_static("create_problem", [](const Initializer& init) { return Setup::CreateProblem(init); }, py::return_value_policy::take_ownership);  // "Creates an instance of the problem identified by name parameter.", py::arg("problemType"), py::arg("prependExoticaNamespace"));
     setup.def_static("create_scene", [](const Initializer& init) { return Setup::CreateScene(init); }, py::return_value_policy::take_ownership);
-    setup.def_static("create_map", [](const Initializer& init) { return Setup::CreateMap(init); }, py::return_value_policy::take_ownership);          // "Creates an instance of the task map identified by name parameter.", py::arg("taskmapType"), py::arg("prependExoticaNamespace"));
     setup.def_static("print_supported_classes", &Setup::PrintSupportedClasses, "Print a list of available plug-ins sorted by class.");
     setup.def_static("get_initializers", &Setup::GetInitializers, py::return_value_policy::copy, "Returns a list of available initializers with all available parameters/arguments.");
     setup.def_static("get_package_path", &ros::package::getPath, "ROS package path resolution.");
--- a/exotica_python/src/pyexotica.cpp
+++ b/exotica_python/src/pyexotica.cpp
@@ -496,6 +496,7 @@ PYBIND11_MODULE(_pyexotica, module)
     setup.def_static("get_dynamics_solvers", &Setup::GetDynamicsSolvers, "Returns a list of available dynamics solvers plug-ins.");
     setup.def_static("create_solver", [](const Initializer& init) { return Setup::CreateSolver(init); }, py::return_value_policy::take_ownership);    // "Creates an instance of the solver identified by name parameter.", py::arg("solverType"), py::arg("prependExoticaNamespace"));
     setup.def_static("create_problem", [](const Initializer& init) { return Setup::CreateProblem(init); }, py::return_value_policy::take_ownership);  // "Creates an instance of the problem identified by name parameter.", py::arg("problemType"), py::arg("prependExoticaNamespace"));
+    setup.def_static("create_scene", [](const Initializer& init) { return Setup::CreateScene(init); }, py::return_value_policy::take_ownership);
     setup.def_static("create_map", [](const Initializer& init) { return Setup::CreateMap(init); }, py::return_value_policy::take_ownership);          // "Creates an instance of the task map identified by name parameter.", py::arg("taskmapType"), py::arg("prependExoticaNamespace"));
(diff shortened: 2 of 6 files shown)


VladimirIvan:
Can you add an example or test for this?

christian-rauch:
I added a test, but I am not sure if it is actually used. E.g. a `self.assertTrue(False)` doesn't raise any error.
I also added the method `get_tree_map` to check which scene frames are defined.

christian-rauch:
The method [`KDL::Segment::setFrameToTip`](https://github.com/orocos/orocos_kinematics_dynamics/blob/eaa92da4bbc2f6156d1f1a895c76ac16fd832a20/orocos_kdl/src/segment.hpp#L161) is only available with KDL 1.4 onwards, which is not available on ROS kinetic.

Should I disable this (via preprocessor macro) for KDL lower than 1.4? Or is there an alternative?

wxmerkt:
How is the newly added method different from using attach object?

exotica_core/include/exotica_core/kinematic_element.h
```diff
@@ -41,6 +41,8 @@ namespace exotica
 class KinematicElement
 {
 public:
+    typedef std::shared_ptr<KinematicElement> Ptr;
```
VladimirIvan: Rename to `KinematicElementPtr` and define it outside of the class.

exotica_core/include/exotica_core/setup.h
```diff
@@ -90,6 +90,12 @@ class Setup : public Object, Uncopyable
         ret->InstantiateInternal(init);
         return ret;
     }
+    static exotica::ScenePtr CreateScene(const Initializer& init)
+    {
+        exotica::ScenePtr ret = std::make_shared<exotica::Scene>();
+        ret->InstantiateInternal(init);
+        return ret;
+    }
```
VladimirIvan: How is this going to be used? We can't assign a Scene created like this to the Problem. Using the Scene outside of a problem opens it to more problems and misuse. Once a Scene is created inside a problem it can be used and modified directly, no need for creating an independent scene.

exotica_python/src/pyexotica.cpp
```diff
@@ -496,7 +496,8 @@ PYBIND11_MODULE(_pyexotica, module)
...
     setup.def_static("create_problem", [](const Initializer& init) { return Setup::CreateProblem(init); }, py::return_value_policy::take_ownership);  // "Creates an instance of the problem identified by name parameter.", py::arg("problemType"), py::arg("prependExoticaNamespace"));
-    setup.def_static("create_map", [](const Initializer& init) { return Setup::CreateMap(init); }, py::return_value_policy::take_ownership);          // "Creates an instance of the task map identified by name parameter.", py::arg("taskmapType"), py::arg("prependExoticaNamespace"));
+    setup.def_static("create_scene", [](const Initializer& init) { return Setup::CreateScene(init); }, py::return_value_policy::take_ownership);
+    setup.def_static("create_map", [](const Initializer& init) { return Setup::CreateMap(init); }, py::return_value_policy::take_ownership);  // "Creates an instance of the task map identified by name parameter.", py::arg("taskmapType"), py::arg("prependExoticaNamespace"));
```
VladimirIvan: I didn't expose `Setup::CreateMap` on purpose. It gives access to creating a map that will need to the scene which is done outside the initializer. Maps created like this won't work without several other steps that are not documented. It's best to create a problem and get the map from there.

exotica_python/src/pyexotica.cpp
```diff
@@ -960,8 +961,17 @@ PYBIND11_MODULE(_pyexotica, module)
...
     scene.def("get_model_state", &Scene::GetModelState);
     scene.def("get_model_state_map", &Scene::GetModelStateMap);
+    scene.def("get_tree_map", [](Scene& scene) {
+        std::map<std::string, std::shared_ptr<KinematicElement>> map_shared;
+        for (const auto& m : scene.GetTreeMap())
+        {
+            map_shared[m.first] = m.second.lock();
+        }
+        return map_shared;
+    });
```
VladimirIvan: The `TreeMap` elements get out of scope after loading a scene. If they get exposed as a shared pointer, the user will end up holding a dangling pointer or even worse, blocking the scene editing process.

Expose only `get_tree_names`, extracting only the keys of the map.

exotica_python/src/pyexotica.cpp
```diff
@@ -1097,6 +1107,10 @@ PYBIND11_MODULE(_pyexotica, module)
         return vec;
     });
 
+    // KinematicElement
+    // TODO: bindings for methods
+    py::class_<KinematicElement, std::shared_ptr<KinematicElement>>(kin, "KinematicElement");
```
VladimirIvan: I prefer to keep this class out of the Python API. It opens up the code to a lot of low level trouble.

exotica_python/src/pyexotica.cpp
```diff
@@ -960,8 +961,17 @@ PYBIND11_MODULE(_pyexotica, module)
...
+        std::map<std::string, std::shared_ptr<KinematicElement>> map_shared;
+        for (const auto& m : scene.GetTreeMap())
+        {
+            map_shared[m.first] = m.second.lock();
+        }
+        return map_shared;
+    });
     scene.def("set_model_state", (void (Scene::*)(Eigen::VectorXdRefConst, double, bool)) & Scene::SetModelState, py::arg("x"), py::arg("t") = 0.0, py::arg("update_trajectory") = false);
     scene.def("set_model_state_map", (void (Scene::*)(std::map<std::string, double>, double, bool)) & Scene::SetModelState, py::arg("x"), py::arg("t") = 0.0, py::arg("update_trajectory") = false);
+    scene.def("set_frame_state", &Scene::SetFrameState);
```
VladimirIvan: `attach_object_local ` already implements this.

exotica_python/src/pyexotica.cpp
wxmerkt: ... and it may change when we turn `KinematicTree` into a plug-in to open up for other kinematics solvers

exotica_core/include/exotica_core/setup.h
christian-rauch: The `create_scene` / `CreateScene` methods are for parsing and initialising a scene from an initialiser. This is for cases where you want to obtain information about the scene (e.g. number and names of links and joints) before creating the problem. While the scene initialiser allows you to change the loaded scene in some ways, there is no way to obtain information about the scene, other than parsing the URDF yourself.
At the moment it is not meant that this scene is reused in a problem. Hence, there are no side-effects of accessing a scene like this.

exotica_python/src/pyexotica.cpp
christian-rauch: The `create_map` binding was there before. It's just highlighted in the diff because the comment that has been shifted by clang-format.

exotica_core/include/exotica_core/setup.h
wxmerkt: I agree there are uses where you just want to get Kinematics from a Scene without a problem. I think this requires documentation (to be added) that

(a) A Scene created via this way can only be used for obtaining kinematics etc. - it won't be assignable nor workable with a problem or solver as it's stand-alone.

exotica_python/src/pyexotica.cpp
wxmerkt: Are you actively using it? And in what way? It is a lot of complexity that can irritate users, and it won't work without a scene. So if it's not used by you, let's remove it being exposed.

christian-rauch: To be precise, `create_map` / `createMap` was added with the first commit that introduced the python wrapper: https://github.com/ipab-slmc/exotica/commit/3e482ea74ee64e34817fb6e5c64147cbf9ff1366 :-)

I am never using `create_map`. If it is never used outside I can also just remove it in this PR. If `Setup::CreateMap` is not intended to be used as part of public API, maybe we should just make it private (or remove it if it is not used at all).

exotica_python/src/pyexotica.cpp
christian-rauch: I exposed it mainly for the `get_tree_map`. If `get_tree_map` changes it's function to just provide the names as get_tree_names, exposing the `KinematicElement` is not required anymore and will be removed.

exotica_python/src/pyexotica.cpp
wxmerkt: Let's remove it then from the public API. There won't be a viable use case.

christian-rauch: `Setup::CreateMap` was only used by `PlanningProblem` once. I declared it therefore as friend class.

exotica_python/src/pyexotica.cpp
christian-rauch: `Scene::AttachObject` and `Scene::AttachObjectLocal` seem to do much more.
`Scene::SetFrameState` was intended to just change the pose of a frame, that is already part of the scene.

With `AttachObject` the behaviour will be inconsistent:
If I add a frame via the scene initialiser, it will not be part of the `attached_objects_`. A `DetachObject` will raise an exception in this case. Once I change the pose with `AttachObject`, a `DetachObject` call will not throw. Also, this will not really detach the object, but attach it with the identity transform to the root frame.

christian-rauch: What is the usage of `attached_objects_` actually? Is it to keep track of the object's pose when updating frames?

In:
https://github.com/ipab-slmc/exotica/blob/59fb6585561f2f5f07930b9a6aa789a1cb2dbd2b/exotica_core/src/scene.cpp#L546-L549
it appears as if all attached objects get attached locally again, with the same name, parent and pose.

VladimirIvan: The attached objects keep track of objects attached to the robot after resetting/reloading the scene. After the scene is re-loaded, the objects gets re-attached to the same objects they were attached to before re-loading. Since you may not have the parent object created yet, you don't want to just blindly attach objects to parents that may be added in the next iteration. It is anly required when you change the parent.

`AttachObjectLocal` does a little extra work because it allows you to change the parent.
It makes the setting of the frame explicit, you always specify the parent by name so that frame you provide is in the parent frame. Unless the time to look up the new parent is too much for you, just use `AttachObjectLocal`. It calls the the same function to update the to-tip-frame.

christian-rauch: Ok, I am using `AttachObjectLocal` now and removed `SetFrameState`.
The scene initiliser has a `AttachLinks` that I can use in place of `Links` to circumvent the inconsistency with `DetachObject`.

exotica_python/src/pyexotica.cpp
```diff
@@ -960,6 +960,14 @@ PYBIND11_MODULE(_pyexotica, module)
     scene.def("get_model_joint_names", &Scene::GetModelJointNames);
     scene.def("get_model_state", &Scene::GetModelState);
     scene.def("get_model_state_map", &Scene::GetModelStateMap);
+    scene.def("get_tree_names", [](Scene& scene) {
+        std::vector<std::string> frame_names;
+        for (const auto& m : scene.GetTreeMap())
+        {
+            frame_names.push_back(m.first);
```
wxmerkt: Let's try to use `emplace_back` in new code.

christian-rauch: `emplace_back` is constructing an object in place. I think in this case, `emplace_back` would just call the string copy constructor, which effectively is the same as `push_back` the object.

christian-rauch:
This PR basically now implements the `Setup::CreateScene` and `Scene::GetTreeMap` (which is used for `get_tree_names`), and removes `create_map`.
Is there anything missing to be merged?

PR #515, opened 2019-02-13
[exotica_core] Update MoveIt planning scene prior in GetScene()

wxmerkt:
This should address what we discussed today regarding experiments.


wxmerkt:
Actually, not quiet. The method only updates the kinetic joints, not when objects have been moved or added to the scene. That is a todo we need to address later.

PR #519, opened 2019-02-21
Adding potential new feature to pyexotica

cmower:
I found it frustrating having to keep stop/starting ros launch in order to tune the rho parameters for task maps. To combat this issue I implemented a script `interactive_rho_tuning.py` that allows you to tune the rho's without having to constantly stop/start. See [here](https://youtu.be/g5hPmweo_FA) for an example of its use. 

If this sort of feature is something you would like in exotica let me know any changes/additions, otherwise feel free to close the pull request. 

# Checklist

- [x] add unit test(s) (NA)

- [x] ensure tests build and check results: run `catkin run_tests exotica && catkin_test_results` (NA)


exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,33 @@
+import Tkinter
```
VladimirIvan: New run dependency not included in the package.xml

VladimirIvan:
dynamic_reconfigure was designed precisely for this purpose. Why use custom UI?

cmower:
@VladimirIvan exotica doesn't broadcast `Rho` values as configurable ros parameters so cannot be picked up by `dynamic_reconfigure`.

wxmerkt:
I like the idea but haven't been able to test it in particular. I see your issues with `dynamic_reconfigure` and like the simplicity of this interface. 

You can get a callback in `dynamic_reconfigure` to extract the values and set them to whatever you want - that could help? [I've only used that in C++ before]

cmower:
> You can get a callback in `dynamic_reconfigure` to extract the values and set them to whatever you want - that could help? [I've only used that in C++ before]

This sounds like a good idea to look into. They list in the ros wiki for `dynamic_reconfigure` that the Python API is stable so I guess should work. 

Two questions:

1. The feature could be generalized to re-configure other parameters on-the-fly (eg. solver params). Would this be worth the effort or should I stick with `Rho` parameters? 

1. Is there a way to discriminate task maps based on their use as either cost terms or hard constraints? What would be useful is to have a function `is_cost_term()` (for example). Having this function would allow me to perform the following

In XML:
```xml
<Cost>
  <Task Task="ACostTaskMap" Rho="1"/>
</Cost>

<Equality>
  <Task Task="AnEqualityTaskMap"/>
</Equality>
```
In Python:
```python
task_maps = problem.get_task_maps()
cost_task_maps = []
for k in task_maps.keys():
  task_map = task_maps[k]
  if task_maps[k].is_cost_term(): 
    # True -> task map used in cost function
    cost_task_maps.append(task_map)

# Now cost_task_maps is a list of 1 element
# i.e. the task map with the name "ACostTaskMap"
```

wxmerkt:
> 1. The feature could be generalized to re-configure other parameters on-the-fly (eg. solver params). Would this be worth the effort or should I stick with `Rho` parameters?

Only implement what you need for your own work right now, it can always be extended later once required.

> 2\. Is there a way to discriminate task maps based on their use as either cost terms or hard constraints? What would be useful is to have a function `is_cost_term()` (for example). Having this function would allow me to perform the following

The `equality`, `inequality`, `cost` private members (of type `Task` or `TimeIndexedTask`) are or should be exposed as read-only for most problem types. You can get the tasks included in them from there - or add a convenience method to `Task`/`TimeIndexedTask`.

wxmerkt:
What is the status of this pull request? It seems like after adding the run dependency it could be good?

cmower:
This hasn't been a priority for me for a while so I didn't look at it till now. 
* I've added a new save feature. See a demo [here](https://youtu.be/K3o06NaaUT0). 
* Simply using a `try-except` solves the error when trying to handle equality/inequality constraints.

exotica_python/src/pyexotica/interactive_rho_tuning.py
cmower: I've added Tkinter to `package.xml` as advised however it is in standard library for Python (see [here](https://docs.python.org/2/library/tk.html)) so was this really necessary?

cmower:
Seems Travis CI build failed. I do not know why?

cmower:
Prepend `python-` to `Tkinter` in `package.xml`. Perhaps this solves Travis CI issue...

exotica_python/package.xml
```diff
@@ -14,4 +14,5 @@
   <depend>shape_msgs</depend>
   <depend>python-matplotlib</depend>
   <depend>python-pyassimp</depend>
+  <depend>python-Tkinter</depend>
```
wxmerkt: No, the correct key is `python-tk`, cf. https://github.com/ros/rosdistro/blob/61959ac7a9d2d2bc91c0555ce139c22de428a917/rosdep/python.yaml#L4255

cmower: resolved, also added `python-rospkg` as per [here](http://docs.ros.org/jade/api/catkin/html/howto/format1/python_module_dependencies.html)

cmower:
- [x] Travis CI passed all checks
- [x] Squashed commits into 1 and force pushed

Waiting for final Travis CI checks to be completed and ready to merge as far as i can tell.

exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import print_function
+import os
+from time import time
+import Tkinter as tk
+import rospkg as rp
+
+__all__=['Tuning']
+
+class Tuning(object):
```
wxmerkt: Can you please select a more descriptive name for this class, e.g., `UnconstrainedEndPoseProblemInteractiveCostTuning` or `InteractiveCostTuning`. As far as I can tell it now only supports costs, not constraints - and only for end-pose problems.

exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import print_function
+import os
+from time import time
+import Tkinter as tk
+import rospkg as rp
+
+__all__=['Tuning']
```
wxmerkt: PEP8 formatting (here and in the example file)

exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Initial setup
+        self.master = tk.Tk()
+        self.problem = problem
+
+        # Set title
+        self.master.winfo_toplevel().title("Rho Tuning")
+
+        # Set icon
+        # note, rp.RosPack().get_path('exotica') throws error
```
wxmerkt: Yep, because the core package is now `exotica_core` and `exotica` is only a metapackage - which does not work with ROS pack. A metapackage is like a stack, i.e., you can use `rosstack`, cf. https://github.com/ros/rospack/issues/26. So your solution is:

```python
rp.RosStack().get_path('exotica')
```

exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,86 @@
...
+        icon = rp.RosPack().get_path('exotica_core')[:-13] + '/exotica/doc/images/EXOTica_icon.png'
```
wxmerkt: I don't like this workaround. It has some assumptions on exotica_core being next to exotica, as in our source builds. This may not hold. I have posted a fix using RosStack above.

exotica_python/src/pyexotica/interactive_rho_tuning.py
```diff
@@ -0,0 +1,86 @@
...
+        tk.Button(self.master, text="Set", command=self.set_button).grid(row=n_task_maps+1, column=0, pady=4)
+        tk.Button(self.master, text="Save", command=self.save_button).grid(row=n_task_maps+1, column=1, pady=4)
+        tk.Button(self.master, text="Quit", command=self.quit_button).grid(row=n_task_maps+2, column=0, pady=4)
+
+    def set_button(self):
+        """Sets rho parameters in entries into Exotica problem."""
+        print("Setting Rho parameters:")
+        for k in self.cost_task_map_names:
+            r = float(self.entries[k].get())
+            self.problem.set_rho(k, r)
```
wxmerkt: Note: this only works for end-pose problems.

cmower:
* This will work for unconstrained/constrained problems however the tuning is only for cost terms. It doesn't make sense for equality constraints and will only make sense for inequality constraints if you want to switch the sign and so essentially switching between <= and =>.
* New math and reset features added, see demo [here](https://youtu.be/F9Zu8y4mZeg). 
* All other points raised above have been addressed in latest commit.

exotica_python/src/pyexotica/interactive_cost_tuning.py
```diff
@@ -4,31 +4,33 @@
...
         # Set icon
-        # note, rp.RosPack().get_path('exotica') throws error
-        icon = rp.RosPack().get_path('exotica_core')[:-13] + '/exotica/doc/images/EXOTica_icon.png'
+        icon = rp.RosStack().get_path('exotica') + '/doc/images/EXOTica_icon.png'
         img = tk.PhotoImage(file=icon)
         self.master.tk.call('wm', 'iconphoto', self.master._w, img)
 
         # Grab current rhos and cost task map names
         self.rho = {}
+        self.origional_rho = {}
```
wxmerkt: Typo

wxmerkt:
Thank you for adding this feature :-)

PR #526, opened 2019-03-08
Cleanup time-indexed problems

wxmerkt:
- Moves getter and setter for tasks to `EndPoseTask` and `TimeIndexedTask`: This code was massively duplicated across cost/equality/inequality across several problems.
- Introduces an abstract class for time-indexed kinematic problems and templates of it by deleting unneeded functionality. This reduces code duplication between similar, specialised classes and also removes the subtle differences and bugs we had between seemingly similar problems.
- Adds a lot of doxygen comments to the API and moves some members from public to private where not required.


wxmerkt:
Is this good to go @VladimirIvan?

PR #528, opened 2019-03-11
distinguish between controlled and model joints and links

christian-rauch:
Since EXOTica has the concept of the model state (the total DOF that describes the state of the kinematic chain) and the controlled state (the subset of the model state that can be controlled), I would like the API to reflect this more via the method names.
When querying the attributes (joints, links) and values (state) of a kinematic chain, the method name should contain which set is queried (i.e. the total model set of the controlled subset).
E.g. when using `GetJointNames` it is unclear if this will return the total set or only the controlled subset of joints.

This breaks public API but I think it's worth to clarify this naming.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,11 @@
+# Changelog
+
+## [Unreleased]
+### Changed
+- renamed KinematicTree::GetJointNames -> KinematicTree::GetControlledJointNames
+- renamed Scene::GetJointNames -> Scene::GetControlledJointNames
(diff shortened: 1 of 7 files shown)


wxmerkt:
@christian-rauch Can you please add a `CHANGELOG.rst` referencing the PR given this is an API breaking change? Thank you! I will merge after

christian-rauch:
I added a `CHANGELOG.rst` to the root.
I used https://keepachangelog.com/en/1.0.0/ as guidance for formatting the changelog.

PR #538, opened 2019-04-02
Chain links

christian-rauch:
This PR adds a new method `GetKinematicChainLinks` which behaves similarly to `GetKinematicChain` but returns the link/frame names instead of the joint names.

This PR also:
- adds a script to apply clang-format conveniently (instead of remembering the command)
- reformats the PR checklist as a comment to not appear in the final PR text but only show to the committer when initially creating the PR

### DIFF
--- a/exotica_core/include/exotica_core/kinematic_tree.h
+++ b/exotica_core/include/exotica_core/kinematic_tree.h
@@ -197,7 +197,14 @@ public:
     /// @param begin link name from which the chain starts
     /// @param end link name at which the chain ends
     /// @return list joints between begin and end
-    std::vector<std::string> GetKinematicChain(const std::string& begin, const std::string& end);
+    std::vector<std::string> GetKinematicChain(const std::string& begin, const std::string& end) const;
+
+    /// @brief GetKinematicChainLinks get the links between begin and end of kinematic chain
+    /// @param begin link name from which the chain starts
+    /// @param end link name at which the chain ends
+    /// @return list link between begin and end
+    std::vector<std::string> GetKinematicChainLinks(const std::string& begin, const std::string& end) const;
+
     void SetModelState(Eigen::VectorXdRefConst x);
     void SetModelState(std::map<std::string, double> x);
     Eigen::VectorXd GetControlledState();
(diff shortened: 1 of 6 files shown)


christian-rauch:
I had to close/reopen the PR because I renamed the wrong function and I wanted to prevent that this gets merged before I fix it.
Read for review now.

christian-rauch:
Ping @wxmerkt @VladimirIvan

exotica_core/include/exotica_core/kinematic_tree.h
```diff
@@ -198,6 +198,13 @@ class KinematicTree : public Uncopyable
     /// @param end link name at which the chain ends
     /// @return list joints between begin and end
     std::vector<std::string> GetKinematicChain(const std::string& begin, const std::string& end);
+
+    /// @brief GetKinematicChainLinks get the links between begin and end of kinematic chain
+    /// @param begin link name from which the chain starts
+    /// @param end link name at which the chain ends
+    /// @return list link between begin and end
+    std::vector<std::string> GetKinematicChainLinks(const std::string& begin, const std::string& end);
```
wxmerkt: `std::vector<std::string> GetKinematicChainLinks(const std::string& begin, const std::string& end) const;`

exotica_core/src/kinematic_tree.cpp
```diff
@@ -1023,6 +1023,33 @@ std::vector<std::string> KinematicTree::GetKinematicChain(const std::string& beg
...
+    {
+        if (!tree_map_.count(l))
+        {
+            ThrowPretty("Link '" + l + "' does not exist.");
+        }
+    }
+
+    // get chain in reverse order, end...begin
+    std::vector<std::string> chain;
+    for (std::shared_ptr<const KinematicElement> l = tree_map_.at(end).lock(); l->segment.getName() != begin; l = l->parent.lock())
```
wxmerkt: Is this `const` required or beneficial?

christian-rauch: It's not required but it prevents accidental changes to the link. Alternative for loops, e.g. `for(const KinematicElement & link : links) { ... }`, use to operate on const references too.

exotica_core/include/exotica_core/kinematic_tree.h
christian-rauch: I made both methods (`GetKinematicChain` and `GetKinematicChainLinks`) const.

PR #542, opened 2019-04-17
[DO NOT MERGE] Revert "[exotica_python] Support Python2 and Python3"

wxmerkt:
Somehow Travis is failing (timing out) since merging the Python3 fix PR - even though it passed Continuous Integration. Which is odd - my suspicion is memory use of Pinocchio or another issue. Nonetheless, I'd like to see whether a revert _would_ in theory fix CI.

**Please do not merge. This is for diagnosis only.**

Reverts ipab-slmc/exotica#541


wxmerkt:
A couple restarts later, Travis went green again so it wasn't the PR.

PR #552, opened 2019-05-06
[exotica_core_task_maps] Update to manipulability task map

cmower:
* User can now specify to compute manipulability measure based on only position or both position and rotation.
* Documented and applied code formatting.

# Checklist

-  [x] code formatting: run `find -name '*.cpp' -o -name '*.h' -o -name '*.hpp' | xargs clang-format-3.9 -style=file -i` in the root directory

- [x] add unit test(s)

- [x] ensure tests build and check results: run `catkin run_tests exotica && catkin_test_results`

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
@@ -52,7 +52,7 @@ namespace exotica
 ///   - derivatives of \f$\Phi\f$ are computed using finite differences.
 ///
 /// Todo
-///   - Update user options, that is allow user to specify to compute based on translation, rotation, all, yoshikawa, or asada as in Peter Corkes' toolbox.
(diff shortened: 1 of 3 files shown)


cmower:
I don't have a specific paper where this is done. However, I know the [`maniplty`](https://github.com/petercorke/robotics-toolbox-matlab/blob/0ca01aac26b4475094845982b9a5e80b02c024fd/%40SerialLink/maniplty.m#L27) function in Peter Corkes' toolbox has this functionality. So I am guessing it is used somewhere. I wanted to check if there was a significant relation with iteration count and number of task space dimensions when using this task map. Unfortunately, for my stuff it wasn't significant enough however I thought it would be useful to have this option in the implementation since I had done it anyway.

cmower:
Seems some checks in Travis CI failed. I'm not sure how to interpret this when i check the details..

wxmerkt:
I like the options in Peter Corke's library (translation, rotation, all, yoshikawa, asada). Perhaps it'd be a good to idea either to implement it that way or to add a TODO/reference for future consideration.

cmower:
@wxmerkt do you have any idea why Travis CI is failing? I haven't changed much in the code. I agree, task map should have options similar to Peter Corkes toolbox but I think it's best to make it a todo. Does a todo go in the code or as an issue?

wxmerkt:
Todo goes in the code in this case. 

The Travis failure is a known issue for the past few days with Travis, cf. https://github.com/ros-industrial/industrial_ci/issues/364.

cmower:
Appended a todo. All seems well to merge?

PR #553, opened 2019-05-07
[exotica_core] Make generate_initializers.py Python3 compatible

wxmerkt:
Fixes use case where the standard Python interpreter is Python 3. Thanks to @Yunaik for discovering and suggesting a fix.

cc: @wenbin-hu for verification


wenbin-hu:
Yes, it works for me. Thanks.

PR #555, opened 2019-05-09
constant initialiser (read-only) for Instantiable

christian-rauch:
The `Initializer` is meant to configure any `Instantiable` object in EXOTica. It is meant to be read-only and this PR declares the `Instantiate` as a method that only reads from the `Initializer`.

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/center_of_mass.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/center_of_mass.h
@@ -44,7 +44,7 @@ public:
     CenterOfMass();
     virtual ~CenterOfMass();
 
-    void Instantiate(CenterOfMassInitializer& init) override;
+    void Instantiate(const CenterOfMassInitializer& init) override;
     void AssignScene(ScenePtr scene) override;
 
     void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi) override;
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/center_of_mass.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/center_of_mass.h
@@ -44,7 +44,6 @@ public:
     CenterOfMass();
     virtual ~CenterOfMass();
 
-    void Instantiate(const CenterOfMassInitializer& init) override;
     void AssignScene(ScenePtr scene) override;
 
     void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi) override;
@@ -64,8 +63,6 @@ private:
     visualization_msgs::Marker goal_marker_;
     bool enable_z_;
     int dim_;
-
-    CenterOfMassInitializer init_;
 };
 }
 
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/collision_check.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/collision_check.h
@@ -42,7 +42,6 @@ public:
     CollisionCheck();
     virtual ~CollisionCheck();
 
-    void Instantiate(const CollisionCheckInitializer& init) override;
     void AssignScene(ScenePtr scene) override;
 
     void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi) override;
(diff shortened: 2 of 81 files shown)


christian-rauch:
We could keep the old method with a `[[deprecated]]` and const cast it to the new method. You could then update your out-of-three code later and remove the deprecation notice once you are done. But then, this deprecation notice might stay for much longer than initially anticipated.

christian-rauch:
What do I need to run to reproduce the test results of the CI?
On my local build, `rostest exotica_examples python_tests.launch` and `catkin run_tests` report success. It's actually difficult to see which tests fail on the CI, there is a lot of output.

wxmerkt:
> What do I need to run to reproduce the test results of the CI?

Are you running `catkin_test_results [.. path to your build folder..]` after `catkin run_tests`?

christian-rauch:
Thanks, I did not know about `catkin_test_results`. I am now getting the same message.

> > What do I need to run to reproduce the test results of the CI?
> 
> Are you running `catkin_test_results [.. path to your build folder..]` after `catkin run_tests`?

Thanks, I did not know about `catkin_test_results`. I am now getting the same message.

christian-rauch:
Would you consider merging this if I add versions of the methods that take the non-const initialiser?
Which methods would that be?

wxmerkt:
> Would you consider merging this if I add versions of the methods that take the non-const initialiser?
Which methods would that be?

The non-const base instantiate should still exist and preserve behaviour, but can throw a deprecated warning. This would avoid having to change all dependent code now.

christian-rauch:
Since the `Instantiate` method is purely virtual, it needs to be implemented in any case.
But I think configuration should be optional, i.e. the `Instantiate` method does not need to be implemented.

I made the new method virtual and the old non-const deprecated:
```C++
// new const initialisers
virtual void Instantiate(const C& init){}

// old (deprecated) non-const initiliasers
    [[deprecated]] virtual void Instantiate(C& init)
{
    Instantiate(init);
}
```

(This strange deprecate indent comes from the clang format style.)

When you compile your problems or task maps with the old non-const initialiser, you should get a warning like:
`warning: 'void exotica::Instantiable<C, <template-parameter-1-2> >::Instantiate(C&) [with C = exotica::JointPriorInitializer; <template-parameter-1-2> = exotica::JointPriorInitializer]' is deprecated [-Wdeprecated-declarations]`

christian-rauch:
This deprecation is not exactly working as intended. I am getting this warning for all objects in the EXOTica code base. I think you need to explicitly `const` cast the initialiser to use the new method and to silence the warning.
Do you still want to keep this warning?

christian-rauch:
By default, the `Instantiate` method now sets the internal parameters:
```C++
virtual void Instantiate(const C& init)
{
    parameters_ = init;
}
```

In the EXOTica code base, we often just set a class member to the same `Initializer` structure used for `Instantiate`. This will remove a lot of boiler plate code that just sets this member.
I.e. if you do not have to do special setups in your solver or task map, you can use the new member `parameters_` without implementing the `Instantiate` method.

wxmerkt:
That is a great feature. Do you mind carrying it through the implementations and removing the instances where it just set a `parameters_`/`init_` so that it is not hiding or duplicating? [In some cases it also was inconsistently named]. If it's too much hassle, we can file an issue for it and do it later.

christian-rauch:
@wxmerkt Yes, maybe we should do this in a separate PR. I just added this here because we changed the method from pure virtual to virtual and this is a more sensible default behaviour for the virtual method.
However, shadowing `parameters_` by a derived class might be an issue. If the CI is ok, would save this change to another issue/PR. If I change the implementation storing the parameters in all solver and task maps, it becomes unrelated to the original intention of this PR.

christian-rauch:
... I removed the Initialisers that have the default behaviour to make sure the solver/tasks behave correctly in the CI tests.

wxmerkt:
Thank you, Christian! Happy to merge once it passes CI :-)

wxmerkt:
Yep, regarding our earlier conversation - if you turn on errors for overloaded virtuals, it throws on the deprecated:

If you add the following to `CMakeLists.txt` in `exotica_core`:
```cpp
add_compile_options(-Wdeprecated-declarations -Woverloaded-virtual -Werror)
```

- and some derived classes had errors. I've fixed those and included in #562

PR #560, opened 2019-05-10
fix CMake warnings

christian-rauch:
Fixes the recent CI [Kdev warnings](http://build.ros.org/job/Kdev__exotica__ubuntu_xenial_amd64/40/warnings2Result/new/):

> catkin_package() DEPENDS on 'Eigen3' but neither 'Eigen3_INCLUDE_DIRS' nor  'Eigen3_LIBRARIES' is defined.
> 
> catkin_package() DEPENDS on 'ompl' but neither 'ompl_INCLUDE_DIRS' nor  'ompl_LIBRARIES' is defined.

(See: https://answers.ros.org/question/253588/ros-kinetic-eigen/?answer=253596#post-id-253596)


wxmerkt:
Is it possible either of them is different between Kinetic and Melodic? I went back and forth on this in the past.

christian-rauch:
I also got them on melodic and this PR let them vanish.
But I am unsure why they only appeared now and not in one of the previous builds...

PR #562, opened 2019-05-11
Deprecate non-const initializer, activate related warning as error, fix random issues

wxmerkt:
<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/CMakeLists.txt
+++ b/exotica_core/CMakeLists.txt
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(exotica_core)
 
+# TODO: Start fixing warnings: - https://stackoverflow.com/a/9862800
+# -pedantic -Wall -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wlogical-op -Wmissing-declarations -Wmissing-include-dirs -Wnoexcept -Wold-style-cast -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wsign-promo -Wstrict-null-sentinel -Wstrict-overflow=5 -Wswitch-default -Wundef -Werror -Wno-unused
+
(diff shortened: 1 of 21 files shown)


exotica_core/include/exotica_core/property.h
```diff
@@ -129,11 +129,6 @@ class Instantiable : public virtual InstantiableBase
         parameters_ = init;
     }
 
-    [[deprecated]] virtual void Instantiate(C& init)
-    {
-        Instantiate(static_cast<const C&>(init));
-    }
```
christian-rauch: Does this mean that we will not be backward compatible and all old problems/tasks need to adapt to the new virtual method?

exotica_core/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(exotica_core)
 
+# TODO: Start fixing warnings: - https://stackoverflow.com/a/9862800
+# -pedantic -Wall -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wlogical-op -Wmissing-declarations -Wmissing-include-dirs -Wnoexcept -Wold-style-cast -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wsign-promo -Wstrict-null-sentinel -Wstrict-overflow=5 -Wswitch-default -Wundef -Werror -Wno-unused
+
+add_compile_options(-Wdeprecated-declarations -Woverloaded-virtual)# -Werror) # -Wall -Wextra) # -pedantic) # -Werror) # can't use -Werror due to MoveIt / ROS-Console
```
christian-rauch: Do we really want to turn deprecation warnings into errors? It basically has the same effect as not to use deprecation attribute, i.e. the compilation will fail because of mismatching API.

wxmerkt: Not sure about deprecation - but for others over time, yes. Also, I haven't yet added the `-Werror` flag.

exotica_core/include/exotica_core/property.h
wxmerkt: Yes, and I carried it through all solvers I have access to and currently in the workspace (OPT++, SNOPT, KNITRO, IPOPT, etc.). If there's anything I have missed, I am happy to fix it if I get a pointer.

PR #566, opened 2019-05-13
Const getter for parameter and problems

christian-rauch:
- changed GetProblem to return a const-const pointer to the problem, I assume we do not need to change the problem via the motion solver public methods since it's a pointer anyway. 
- added a const getter for parameters
- enabled all warnings as errors (excluding some non-critical stuff) in exotica core

### DIFF
--- a/exotica_core/include/exotica_core/property.h
+++ b/exotica_core/include/exotica_core/property.h
@@ -129,6 +129,7 @@ public:
         parameters_ = init;
     }
 
+    const C& GetParameters() const { return parameters_; }
 protected:
     C parameters_;
 };
--- a/exotica_core/src/task_map.cpp
+++ b/exotica_core/src/task_map.cpp
@@ -73,6 +73,11 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
     if (jacobian.rows() != TaskSpaceDim() && jacobian.cols() != x.rows())
         ThrowNamed("Jacobian dimension mismatch!");
 
+    if (scene_ == nullptr)
+    {
+        ThrowNamed("Scene is not initialised!");
+    }
+
     // Set constants
     constexpr double h = 1e-6;
     constexpr double h_inverse = 1.0 / h;
--- a/exotations/exotica_core_task_maps/src/center_of_mass.cpp
+++ b/exotations/exotica_core_task_maps/src/center_of_mass.cpp
@@ -168,9 +168,9 @@ void CenterOfMass::Initialize()
             {
                 ThrowNamed("Requesting CenterOfMass frame with base other than root! '" << frames_[i].frame_A_link_name << "'");
             }
-            frames_[i].frame_A_link_name = scene_->GetKinematicTree().GetTreeMap()[frames_[i].frame_A_link_name].lock()->segment.getName();
-            frames_[i].frame_A_offset.p = scene_->GetKinematicTree().GetTreeMap()[frames_[i].frame_A_link_name].lock()->segment.getInertia().getCOG();
-            mass_(i) = scene_->GetKinematicTree().GetTreeMap()[frames_[i].frame_A_link_name].lock()->segment.getInertia().getMass();
+            frames_[i].frame_A_link_name = scene_->GetKinematicTree().GetTreeMap().at(frames_[i].frame_A_link_name).lock()->segment.getName();
+            frames_[i].frame_A_offset.p = scene_->GetKinematicTree().GetTreeMap().at(frames_[i].frame_A_link_name).lock()->segment.getInertia().getCOG();
+            mass_(i) = scene_->GetKinematicTree().GetTreeMap().at(frames_[i].frame_A_link_name).lock()->segment.getInertia().getMass();
         }
     }
 
--- a/exotica_core/CMakeLists.txt
+++ b/exotica_core/CMakeLists.txt
@@ -1,11 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(exotica_core)
 
-# TODO: Start fixing warnings: - https://stackoverflow.com/a/9862800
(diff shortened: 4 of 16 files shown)


exotica_core/include/exotica_core/property.h
```diff
@@ -129,6 +129,7 @@ class Instantiable : public virtual InstantiableBase
         parameters_ = init;
     }
 
+    const C& getParameters() const { return parameters_; }
```
wxmerkt: `GetParameters`

christian-rauch:
Wait... I changed `get_problem` from a method to a variable. Going to fix this now.

christian-rauch:
Btw, debugging the tests is sometimes difficult.
E.g. if the test-log says:
```
Unable to find test results for nosetests-test.run_tests.py.xml, test did not run.
Expected results in /home/christian/Development/mpe_base_ws/build/exotica_examples/test_results/exotica_examples/nosetests-test.run_tests.py.xml
```

It actually meant to say:
`TypeError: 'pyexotica._pyexotica.Problems.UnconstrainedEndPoseProblem' object is not callable` :-)

christian-rauch:
ping @wxmerkt

wxmerkt:
I am looking to get a comment from @VladimirIvan on the remove of `py::return_value_policy::reference_internal` - from my understanding it doesn't make a difference as it's the default behaviour since pybind11 v2.0-rc1 but since this is a pointer, I am not 100% sure.

christian-rauch:
Why was it defined as an internal reference in the first place?
`GetProblem` was returning a copy of a shared pointer.

VladimirIvan:
It was an attempt at making python delete the child objects when the parent is deleted. I don't think it should affect `shared_ptr` though. Those should be handled like normal variables and only delete data when the last reference is gone. I think it can be removed but I ultimately added those upon your requests when you both had problems with python leaking memory. The documentation for that policy is [here](https://pybind11.readthedocs.io/en/stable/advanced/functions.html?highlight=return_value_policy).

Have you ran the examples after changing `GetProblem` to `const`? Python uses that heavily. If the returned object is `const`, you can only call `const` methods on it, e.g. you shouldn't be able to run update manually, among others.

christian-rauch:
At least the `unique_ptr` is used as a regular variable in: https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html

Regarding the `const`nes of `get_problem`, I can do the following:
```Python
# setup solver and problem
self.solver = exo.Setup.create_solver(self.sol_conf)
self.problem = exo.Setup.create_problem(self.prob_conf)
self.solver.specify_problem(self.problem)

# try to reset 'number_of_problem_updates_'
prob = self.solver.get_problem()
prob.reset_number_of_problem_updates()
```
I am not sure what should happen, there are no issues or warnings shown. Maybe Python gets a copy after all and it does not change the original problem. I haven't tested this yet.

christian-rauch:
This PR now also contains a pointer check for accessing the scene for numerical gradient computation.

exotica_core/src/task_map.cpp
```diff
@@ -73,6 +73,11 @@ void TaskMap::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef Phi, Eigen::M
     if (jacobian.rows() != TaskSpaceDim() && jacobian.cols() != x.rows())
         ThrowNamed("Jacobian dimension mismatch!");
 
+    if (scene_ == nullptr)
```
wxmerkt: Is this for the case where you create a task map manually without a problem? Or a problem without a proper scene? I am trying to understand how this could be triggered.

christian-rauch: It's triggered if you overwrite `AssignScene` without properly implementing it. It shouldn't happen that often but it helps to debug the problem.

christian-rauch:
I added some const methods for accessing the `KinematicTree`. Where pointers or references are returned, I created an additional `const` method since I am not sure if the returned parts of the kinematic tree are supposed to be changed (there is no documentation about this).

I am adding these commits to this PR because I did not want to create and manage/rebase another branch. Let me know if I shall create a new PR for these unrelated changes, e.g. if the original PR needs more discussion.

wxmerkt:
I am a bit confused - how would the following work (isn't one hidden by the other - how would the compiler decide to throw/not throw?):

```cpp
std::vector<std::weak_ptr<KinematicElement>> GetTree() { return tree_; }
std::vector<std::weak_ptr<KinematicElement>> GetTree() const { return tree_; }
```

Same return, same method name, same arguments - one const, the other non-const - how does one not hide the other?

christian-rauch:
> I am a bit confused - how would the following work (isn't one hidden by the other - how would the compiler decide to throw/not throw?):
> 
> ```c++
> std::vector<std::weak_ptr<KinematicElement>> GetTree() { return tree_; }
> std::vector<std::weak_ptr<KinematicElement>> GetTree() const { return tree_; }
> ```
> 
> Same return, same method name, same arguments - one const, the other non-const - how does one not hide the other?

The compiler decides depending on the `const`-nes of the object on which the method is called: https://stackoverflow.com/questions/17601121/can-we-achieve-polymorphism-through-const-function

However, I am not certain what happens to the pointer returned by a `const` method. The method itself cannot change its object, but you can still change the object through the pointers.

wxmerkt:
To me the non-const one is equivalent to the const one (but I may very well misunderstand things here) - returning an internally stored vector is a const-operation on the class object (the getter isn't changing the `tree_` before returning). Returning a `std::weak_ptr<const ...>` would be different.

christian-rauch:
You are right. Since we are returning a copy of `tree_`, the method `GetTree()` does not change its object. But it's still better to mark it `const` so it can be called on a const object. `GetTree()` should be removed since it does not provide more access options that its `const` counterpart.

However, you can still change the tree elements through the returned pointer and this might be intended.

wxmerkt:
> However, you can still change the tree elements through the returned pointer and this might be intended.

I agree. For now we can mark it as const, then later figure out if a method using `GetTree` actually modifies the tree (which it shouldn't), add a `GetConstTree`, switch usage over, and remove `GetTree`.

christian-rauch:
Added fix for #568

christian-rauch:
I am getting `Jacobian error out of bounds` on the CI from time to time, which can be "fixed" by rerunning the CI job. Should we raise the bounds for this?

wxmerkt:
I've been raising the bounds several times recently - but it still comes up. We may have to raise them again, yes.

Issue #571, opened 2019-05-17
Quadrotor and floating-base dynamics problems deactivated


wxmerkt:
Directly coming from #570, in order to get a fix for the start-state-size in for @include4eto, I will deactivate the Quadrotor solver. The issue is that the floating base state in the kinematics do not match the ones in the dynamics.

We shall come back to provide a fix later.

wxmerkt:
This should be fixed now.

Issue #570, opened 2019-05-17
Handle floating-bases properly


wxmerkt:
Now that we are implementing dynamic problems, the incorrect handling of the floating-base is coming to bite us - it is not straight-forward to extract and propagate the sizes of the position and velocity vectors. However, this would likely mean redesigning the kinematics entirely - which is quite an undertaking.

- [ ] Move from `KDL::Frame` to `Eigen::Isometry3d` in all APIs, i.e., push `KDL::Frame` to within `KinematicTree`.
- [ ] Properly represent the SE(3)/SE(2) floating-base joints in KinematicTree. This may require changing to another kinematics library.
- [ ] Propagate `actuated` / `model` (includes fixed) / `controlled` sizes for position and velocity from a single place

PR #572, opened 2019-05-17
Support full start state (position + velocity) for dynamic problems

wxmerkt:
Required quite a few changes:

- `DynamicsSolver` is now instantiable and owned by the `Scene`, not the problem.
- Deactivated quadrotor and floating-base dynamic problems as we aren't treating floating bases properly (#570, #571)

cc: @include4eto

### DIFF
--- a/exotica_core/include/exotica_core/planning_problem.h
+++ b/exotica_core/include/exotica_core/planning_problem.h
@@ -35,6 +35,7 @@
 #include <string>
 #include <vector>
 
+#include <exotica_core/dynamics_solver.h>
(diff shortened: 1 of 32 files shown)


exotica_core/include/exotica_core/planning_problem.h
```diff
@@ -35,6 +35,7 @@
 #include <string>
 #include <vector>
 
+#include <exotica_core/dynamics_solver.h>
```
VladimirIvan: Shouldn't this have moved to the scene?

PR #580, opened 2019-05-27
[task_maps] Updated LookAt.

cmower:
* Now user only required to specify one `LookAtTarget` frame per end-effector (as in example xml).
* Documentation, test script, and example scripts updated accordingly.
* Removed unnecessary function `get_look_at_target_in_world` from `task_map_py`.

# Checklist

- [x] updated unit test(s)

- [x] ensure tests build and check results: run `catkin run_tests exotica && catkin_test_results`

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/look_at.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/look_at.h
@@ -43,11 +43,6 @@ namespace exotica
 /// \brief Points end-effector to look at a given target.
 /// Looks at a target point by penalizing the vector which defines the orthogonal projection onto a defined line in the end-effector frame.
 ///
-/// The task map relies on defining three frames for each target to be looked at:
-///  - (0) the EffPoint in the end-effector frame (use in task map)
-///  - (1) the LookAtTarget in the end-effector frame (use in task map)
-///  - (2) the LookAtTarget in the world frame (use by user to easily return target in world coordinates - can be retrieved via get_look_at_target_in_world)
-///
 /// \image html taskmap_lookat.png "LookAt task map." width=500px
 ///
 /// Given the point \f$p\f$ (the point to look at) defined in the end-effector space the task map is expressed by
@@ -71,8 +66,6 @@ public:
     void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi) override;
     void Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::MatrixXdRef jacobian) override;
     int TaskSpaceDim() override;
-
-    Eigen::Vector3d get_look_at_target_in_world(const int& i);
 
 private:
     int n_end_effs_;  ///< Number of end-effectors.
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/look_at.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/look_at.h
@@ -40,7 +40,7 @@ namespace exotica
 ///
 /// \ingroup TaskMap
 ///
-/// \brief Points end-effector to look at a given target.
+/// \brief Points end-effector to look at a given target by aligning end-effector z-axis with the target.
 /// Looks at a target point by penalizing the vector which defines the orthogonal projection onto a defined line in the end-effector frame.
 ///
 /// \image html taskmap_lookat.png "LookAt task map." width=500px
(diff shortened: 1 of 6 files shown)


exotations/exotica_core_task_maps/src/look_at.cpp
```diff
@@ -36,30 +36,15 @@ namespace exotica
...
-        Eigen::Vector3d c = Eigen::Map<Eigen::Vector3d>(kinematics[0].Phi(end_effector_id).p.data);      // EffPoint | Eff frame
-        Eigen::Vector3d p = Eigen::Map<Eigen::Vector3d>(kinematics[0].Phi(end_effector_id + 1).p.data);  // LookAtTarget | Eff frame
-
-        // Compute orthogonal orthogonal projection a onto line e->c
-        double alpha = p.dot(c) / c.squaredNorm();
-        Eigen::Vector3d a = alpha * c;
-
-        // Set Phi
-        phi.segment<3>(end_effector_id) = a - p;
+        Eigen::Vector3d p = Eigen::Map<Eigen::Vector3d>(kinematics[0].Phi(i).p.data);
```
wxmerkt: From this, it tells me that now the equation simplifies to be the same same as `EffPosition`, but only considering x and y. Is this correct?

exotations/exotica_core_task_maps/src/look_at.cpp
```diff
@@ -70,47 +55,22 @@ void LookAt::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::Ma
...
-
-    // Verify that the second and third frames are the same
-    for (int i = 0; i < n_end_effs_; ++i)
-    {
-        if (frames_[i + 1].frame_A_link_name != frames_[i + 2].frame_A_link_name)
-        {
-            ThrowPretty("The second and third links (LookAtTarget) need to be the same! Got: " << frames_[i + 1].frame_A_link_name << " and " << frames_[i + 2].frame_A_link_name);
-        }
-    }
+    n_end_effs_ = frames_.size();
```
wxmerkt: You now need to add the following line when overriding Instantiate: `parameters_ = init;`. `parameters_` is inherited for all instantiable objects to store the initializer.

exotations/exotica_core_task_maps/src/look_at.cpp
```diff
@@ -70,47 +55,22 @@ void LookAt::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::Ma
...
+    n_ = n_end_effs_ * 3;
```
wxmerkt: Technically it could be times two - and above instead of setting it to zero for the z-dimension, just omit it. Same for Jacobian - you can just get top-2 instead of top-3.

exotations/exotica_core_task_maps/include/exotica_core_task_maps/look_at.h
```diff
@@ -43,11 +43,6 @@ namespace exotica
 /// \brief Points end-effector to look at a given target.
 /// Looks at a target point by penalizing the vector which defines the orthogonal projection onto a defined line in the end-effector frame.
 ///
```
VladimirIvan: Mention here that the Z axis will align to look at the target.

cmower:
* Points above addressed in latest commit. 
* @wxmerkt I think `LookAt` is still appropriate despite it suffering from the fact that it will look in the opposite direction given a bad initialization or certain trajectories of the `LookAtTarget`. After all, intuitively the end-effector "looks" at the given target point. A `Gaze` constraint could be implemented, i.e. like a virtual cone attached to the end-effector that constrains the motion of robot such that the `GazeTarget` stays within the cone (essentially a virtual fixture). I may implement this in the future.

exotations/exotica_core_task_maps/src/look_at.cpp
```diff
@@ -36,85 +36,29 @@ namespace exotica
...
-            ThrowPretty("The second and third links (LookAtTarget) need to be the same! Got: " << frames_[i + 1].frame_A_link_name << " and " << frames_[i + 2].frame_A_link_name);
-        }
+            jacobian.middleRows<2>(i).col(j) = kinematics[0].jacobian[i].data.topRows<2>().col(j);
     }
 }
 
 int LookAt::TaskSpaceDim()
 {
-    return n_;
+    return 3 * frames_.size() - frames_.size();
```
wxmerkt: `2 * frames_size()`?

exotations/exotica_core_task_maps/src/look_at.cpp
```diff
@@ -36,85 +36,29 @@ namespace exotica
...
 }
 }
```
wxmerkt: `}  // namespace exotica`

wxmerkt:
Thank you, I like the simplicity of it now. One minor change request inline (3x-x=2x), otherwise good to go.

exotations/exotica_core_task_maps/src/look_at.cpp
cmower: 3x - x = 2x

cmower:
Rebased, I think all should be good to merge here.

wxmerkt:
Thank you, will merge once CI goes green.

PR #583, opened 2019-05-29
Hold onto initializer in the planning problem.

cmower:
* Keeping track of the initializer cleans up `interactive_cost_tuning.py` code (see updated file).
* I have another use for this addition in my work.  

<!--
# Checklist

- add unit test(s)

--!>


wxmerkt:
I think there is a better way of doing this: the instantiation order is `InstantiableBase` of the abstract/base class (here `PlanningProblem`) and then `Instantiate` of the particular class. The `Instantiate` of the derived class hereby by default stores the initializer as `parameters_` which can be retrieved using `GetParameters`:

https://github.com/ipab-slmc/exotica/blob/ecd59b298c80fc2ec8d74fc7dd4c59afaf3cd374/exotica_core/include/exotica_core/property.h#L132

Exposing this would be sufficient to get you want you want without adding another copy into `PlanningProblem`.

cmower:
I see what you mean, but I am at a loss of how to expose `GetParameters` to Python?

wxmerkt:
Let's get @VladimirIvan's input - Instantiable is a templated-class type and as far as I know pybind11 only supports fully specified templates to be exposed - but perhaps he knows a way around this.

cmower:
Rebased and force pushed.

cmower:
Despite that mentioned in https://github.com/ipab-slmc/exotica/pull/584#issuecomment-497661455, this is a potential use-case and solution. However, how much do we really care about the cleanliness of `interactive_cost_tuning.py` when it works just fine without this addition?

@VladimirIvan @wxmerkt I am happy to close this PR if you guys are, this may just be creating unnecessary work.

wxmerkt:
If it's just to reduce complexity in one example, I'd prefer not merging this in until we find a cleaner solution.

cmower:
Agreed, will close PR and open later if necessary.

PR #584, opened 2019-05-29
[exotica_core/python] Retrieval method for problems.

cmower:
* This is a solution to problems being discussed in #583 
* Note, cannot directly call `GetParameters` since the type it returns cannot be converted to a Python data type. `GetParametersAsInitializer` gets round this issue. 
* The downside to this solution is that `get_parameters` has to be added in `pyexotica.cpp` every time a new problem is developed. 

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/include/exotica_core/property.h
+++ b/exotica_core/include/exotica_core/property.h
@@ -129,6 +129,11 @@ public:
         parameters_ = init;
     }
 
+    Initializer GetParametersAsInitializer()
+    {
+        return parameters_;
+    }
+
     const C& GetParameters() const { return parameters_; }
 protected:
     C parameters_;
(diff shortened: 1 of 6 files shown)


exotica_core/include/exotica_core/property.h
```diff
@@ -129,6 +129,11 @@ class Instantiable : public virtual InstantiableBase
         parameters_ = init;
     }
 
+    Initializer GetParametersAsInitializer()
+    {
+        return parameters_;
```
wxmerkt: Doesn't this require a cast?

cmower: No, this works as expected.

cmower:
Rebased and force pushed.

cmower:
Since `exo.Initializers.load_xml_full` exists this functionality is somewhat redundant for me now. There may be a use-case for it in some application and I think its worth allowing users to access problem parameters on-the-fly (and task map/solver/scene parameters for that matter). However, this above is not an elegant solution that requires yet more things to remember to implement whenever you want to add new problem scripts. 

Consequently, I think its worth closing this PR and coming back later if a use case comes up.

PR #585, opened 2019-05-30
[exotica_core/python] Load initializer methods.

cmower:
* Two new methods to load problem and solver initializer from xml.
* Exposed both methods to Python.
* I have a case where I want to load a base initializer from xml and update on-the-fly, creating a new problem as I go.

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/include/exotica_core/loaders/xml_loader.h
+++ b/exotica_core/include/exotica_core/loaders/xml_loader.h
@@ -62,6 +62,20 @@ public:
         return Instance()->LoadXML(file_name, parsePathAsXML);
     }
 
+    static Initializer LoadProblemInitializer(const std::string& file_name)
+    {
+        Initializer solver, problem;
+        XMLLoader::Load(file_name, solver, problem);
+        return problem;
+    }
+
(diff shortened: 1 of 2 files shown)


wxmerkt:
Does`LoadInitializer`/`LoadXML`/`LoadXMLFull` not cover your use case?

cmower:
`LoadInitializer` doesn't exist in `xml_loader.h`. However, I want the two distinguishing methods to load the problem and solver initializers individually (as in commit).

cmower:
Rebased and force pushed.

wxmerkt:
Check this out:

https://github.com/ipab-slmc/exotica/blob/a9e3f2406ce3102c468a205b72c835c44bc18e3f/exotica_examples/scripts/example_dynamic_time_indexed_problem#L13-L14

and:

https://github.com/ipab-slmc/exotica/blob/a9e3f2406ce3102c468a205b72c835c44bc18e3f/exotica_examples/test/test_core#L36-L37

I think this covers your use case?

cmower:
Absolutely correct, don't know why i didn't think to look in `Initializers`. Thanks for pointing those methods out!

PR #588, opened 2019-06-03
[examples] Adding forgotten dependency for GazeAtConstraint example.

cmower:
* Forgot to add `exotica_scipy_solver` as a dependency after adding `GazeAtConstraint` example.

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_examples/package.xml
+++ b/exotica_examples/package.xml
@@ -19,6 +19,7 @@
   <depend>sensor_msgs</depend>
   <exec_depend>exotica_collision_scene_fcl</exec_depend>
   <exec_depend>exotica_ompl_solver</exec_depend>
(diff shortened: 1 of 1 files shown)


exotica_examples/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ find_package(catkin REQUIRED COMPONENTS
   exotica_aico_solver
   exotica_core_task_maps
   exotica_ik_solver
+  exotica_scipy_solver
```
wxmerkt: It's not compiling, so not necessary here.

cmower: Reverted `CMakeLists.txt`, squashed, and force pushed.

PR #592, opened 2019-06-09
DDP Dynamics Solver

include4eto:
Adds the following:

- [various] Second-order dynamics derivatives using `Eigen::Tensor` from `Eigen-unsupported`. https://eigen.tuxfamily.org/dox-devel/unsupported/eigen_tensors.html . Set to 0 to all solvers except `CartpoleDynamicsSolver`.
- [`exotica_ddp_solver`] Analytic DDP Solver (Mayne, D. Q. (1966)).
- [`exotica_ddp_solver`] Control-Limited DDP Solver (Tassa et. al., 2014).
- [`exotica_core`] Adds second-order cost function derivatives to `DynamicTimeIndexedProblem`.
- [`exotica_core`] An inlined `BoxQP` header-only quadratic programming solver. See the above Tassa paper for implementation details. 

Changes the following:
- [`exotica_examples`] Re-organize dynamic time indexed examples in a separate folder and add a single script that selects one via command-line arguments.
- [`exotica_core`] `pyexotica` to include `BoxQP`
- [`exotica_core`] `MatrixToTensor` and `TensorToMatrix` conversion tools.


exotica_core/include/exotica_core/tools/box_qp.h
```diff
@@ -0,0 +1,157 @@
...
+{
+typedef struct BoxQPSolution
+{
+    Eigen::MatrixXd Hff_inv;
+    Eigen::MatrixXd x;
+    std::vector<int> free_idx;
+    std::vector<int> clamped_idx;
+} BoxQPSolution;
+
+inline BoxQPSolution BoxQP(Eigen::MatrixXd H, Eigen::VectorXd q,
```
wxmerkt: Is there a reason not use const-ref for these parameters?

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+                   dt_squared;
+            Quu += Eigen::TensorToMatrix((Eigen::Tensor<double, 2>)dynamics_solver_->fuu(x, u).contract(Vx_tensor, dims),
+                                         NU, NU) *
+                   dt_squared;
+            Qux += Eigen::TensorToMatrix((Eigen::Tensor<double, 2>)dynamics_solver_->fxu(x, u).contract(Vx_tensor, dims),
+                                         NU, N) *
+                   dt_squared;
+        }
+
+        Quu_inv = (Eigen::MatrixXd::Identity(Quu.rows(), Quu.cols()) *
```
wxmerkt: Why are these each on their own lines? Somehow the formatting got totally messed up. If you have a preferred format yourself, just turn off clang ``// clang-format off`` and then on again ``// clang-format on``.

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+                       1e-5 +
```
wxmerkt: Maybe add a comment that this is required for numerical stability

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+        Qu = prob_->GetControlCostJacobian(t) + fu.transpose() * Vx;
+        Qxx = prob_->GetStateCostHessian(t) + fx.transpose() * Vxx * fx;
+        Quu = prob_->GetControlCostHessian() + fu.transpose() * Vxx * fu;
+
+        // NOTE: Qux = Qxu for all robotics systems I have seen
+        //  this might need to be changed later on
+        // WARNING_NAMED("ControlLimitedDDPSolver", "Assuming fux = fxu. Make sure this is valid for your robot!");
+        Qux = prob_->GetStateControlCostHessian() + fu.transpose() * Vxx * fx;
+
+        if (parameters_.UseSecondOrderDynamics)
```
wxmerkt: I am trying to think whether it makes sense to have a full if else - and not just add, but the whole equations on their own. Reason being that Eigen/compiler may do some optimisation and it's also easier to read. I.e., if (secondorderdynamics), the whole equations for Qxx, Quu, Qux, else - reduced equations as above.

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+                        fu = dynamics_solver_->fu(x, u);
+
+        fx = fx * dynamics_solver_->get_dt() + Eigen::MatrixXd::Identity(fx.rows(), fx.cols());
+        fu = fu * dynamics_solver_->get_dt();
+
+        // Q = prob_->GetStateCost(t) + prob_->GetControlCost(t) + V; // l + v
+        Qx = prob_->GetStateCostJacobian(t) + fx.transpose() * Vx;  // lx + fx @ Vx
+        Qu = prob_->GetControlCostJacobian(t) + fu.transpose() * Vx;
+        Qxx = prob_->GetStateCostHessian(t) + fx.transpose() * Vxx * fx;
+        Quu = prob_->GetControlCostHessian() + fu.transpose() * Vxx * fu;
```
wxmerkt: You could use a temporary here for fx^T * Vxx - and then only multiplies fx or fu as required. That can help (Croccodyl does it this way).

wxmerkt:
It is failing on melodic on the buildserver - some dependencies are not included and apparently there are some other compilation errors as well.

exotations/dynamics_solvers/exotica_quadrotor_dynamics_solver/src/quadrotor_dynamics_solver.cpp
```diff
@@ -108,4 +108,25 @@ Eigen::VectorXd QuadrotorDynamicsSolver::GetPosition(Eigen::VectorXdRefConst x_i
     return xyz_rpy;
 }
 
+Eigen::Tensor<double, 3> QuadrotorDynamicsSolver::fxx(const StateVector& x, const ControlVector& u)
```
VladimirIvan: Add a TODO comment for the whole class or maybe even throw an 'NotYetImplemented' exception when used.

exotations/dynamics_solvers/exotica_pinocchio_dynamics_solver/src/pinocchio_dynamics_solver.cpp
```diff
@@ -80,4 +80,25 @@ Eigen::MatrixXd PinocchioDynamicsSolver::fu(const StateVector& x, const ControlV
     // return B_;
 }
 
+Eigen::Tensor<double, 3> PinocchioDynamicsSolver::fxx(const StateVector& x, const ControlVector& u)
```
VladimirIvan: Add a TODO comment for the whole class or maybe even throw an 'NotYetImplemented' exception when used.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,217 @@
...
+            break;
+        }
+
+        if (last_cost - current_cost < base_parameters_.FunctionTolerance && last_cost - current_cost > 0)
+        {
+            if (debug_) HIGHLIGHT_NAMED("DDPSolver", "Function tolerance reached.");
+            break;
+        }
+
+        if (iteration == GetNumberOfMaxIterations() - 1 && debug_) HIGHLIGHT_NAMED("DDPSolver", "Max iterations reached.");
```
VladimirIvan: Put `debug_` at the beginning of the condition for quicker lazy evaluation.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,217 @@
...
+        last_cost = current_cost;
+        for (int t = 0; t < T - 1; ++t)
+            prob_->Update(new_U.col(t), t);
+        prob_->SetCostEvolution(iteration, current_cost);
+    }
+
+    // store the best solution found over all iterations
+    for (int t = 0; t < T - 1; ++t)
+    {
+        solution.row(t) = global_best_U.col(t);
```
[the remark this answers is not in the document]
VladimirIvan: Be careful with these. Eigen does not pick these things up and it will iterate over a column like it was a row and easily copy data from/to the wrong memory. Even if it doesn't segfault it will copy garbage and it's very difficult to debug. Go over the all of your files and double check row/column assignments!

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,217 @@
...
+        // eq. 12
+        Eigen::VectorXd delta_uk = k_gains_[t] + K_gains_[t] * dynamics_solver_->StateDelta(prob_->get_X(t), ref_trajectory.col(t));
+        u.noalias() += alpha * delta_uk;
+        // clamp controls
+        if (control_limits.size() == u.size())
+            u = u.cwiseMax(-control_limits).cwiseMin(control_limits);
+        else if (control_limits.size() == 1)
+            u = u.unaryExpr([&control_limits](double x) -> double {
+                return std::min(std::max(x, -control_limits(0)), control_limits(0));
+            });
```
VladimirIvan: It may be better to require `control_limits` to be the same size as `u`. Ideally directly in the dynamics solver. Same for joint and velocity limits if they are used. Then this case won't be required.

exotations/solvers/exotica_ddp_solver/src/control_limited_ddp_solver.cpp
```diff
@@ -0,0 +1,130 @@
...
+void ControlLimitedDDPSolver::Instantiate(const ControlLimitedDDPSolverInitializer& init)
+{
+    parameters_ = init;
+    base_parameters_ = parameters_;
+}
+
+void ControlLimitedDDPSolver::BackwardPass()
+{
+    constexpr double min_clamp_ = -1e10;
+    constexpr double max_clamp_ = 1e10;
```
VladimirIvan: If this is just a large number, use numeric limits.

exotica_core/include/exotica_core/tools/box_qp.h
```diff
@@ -0,0 +1,157 @@
...
+                           Eigen::VectorXd b_low, Eigen::VectorXd b_high, Eigen::VectorXd x_init, const double gamma,
+                           const int max_iterations, const double epsilon)
+{
+    int it = 0;
+    Eigen::VectorXd delta_xf, x = x_init;
+    std::vector<int> clamped, free;
```
VladimirIvan: `free` is a keyword. Use another name.

exotations/dynamics_solvers/exotica_cartpole_dynamics_solver/src/cartpole_dynamics_solver.cpp
```diff
@@ -111,6 +111,86 @@ Eigen::MatrixXd CartpoleDynamicsSolver::fu(const StateVector& x, const ControlVe
...
+                     0,
+                     4 * l_ * l_ * m_p_ * m_p_ * tdot * sin_theta * sin_theta * cos_theta * cos_theta / pow(l_ * m_c_ + l_ * m_p_ * sin_theta * sin_theta, 2) + 2 * l_ * m_p_ * tdot * sin_theta * sin_theta / (l_ * m_c_ + l_ * m_p_ * sin_theta * sin_theta) - 2 * l_ * m_p_ * tdot * cos_theta * cos_theta / (l_ * m_c_ + l_ * m_p_ * sin_theta * sin_theta)}},
+                   {{0, 0, 0, 0},
+                    {0, 0, 0, 0},
+                    {0, 0, 0, 0},
+                    {0, 0, 0, 0}},
+                   {{0, 0, 0, 0},
+                    {0, 0, 0, 0},
+                    {0,
+                     -4 * l_ * m_p_ * m_p_ * tdot * sin_theta * sin_theta * cos_theta / pow(m_c_ + m_p_ * sin_theta * sin_theta, 2) + 2 * l_ * m_p_ * tdot * cos_theta / (m_c_ + m_p_ * sin_theta * sin_theta),
```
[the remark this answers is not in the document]
include4eto: So.. Why not
`using namespace std`?

or, for that matter:
`using namespace Eigen`?

exotations/dynamics_solvers/exotica_double_integrator_dynamics_solver/include/exotica_double_integrator_dynamics_solver/double_integrator_dynamics_solver.h
```diff
@@ -48,6 +48,10 @@ class DoubleIntegratorDynamicsSolver : public DynamicsSolver, public Instantiabl
     StateDerivative fx(const StateVector& x, const ControlVector& u) override;
     ControlDerivative fu(const StateVector& x, const ControlVector& u) override;
 
+    Eigen::Tensor<double, 3> fxx(const StateVector& x, const ControlVector& u) override;
```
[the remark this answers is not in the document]
include4eto: Where do we set the dimensions, though? The base constructor is called before and `AssignScene` is overriden. Plus it shouldn't ever be zero (unless specified).

And the template parameters are `Eigen::Dynamic` and we can't instantiate tensors with it. We can of course do a check every time `fxx` etc. are called and instantiate if not done already.

exotations/dynamics_solvers/exotica_cartpole_dynamics_solver/src/cartpole_dynamics_solver.cpp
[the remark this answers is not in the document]
VladimirIvan: `using namespace std` includes too much and the class/function names are often very common and can easily conflict with/hide other code. It is recommended to use `std::` or do more 'use' only very specialised parts of std.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
include4eto: We want to break regardless of whether `debug_` is true.

VladimirIvan: `if (debug_ && iteration == GetNumberOfMaxIterations() - 1)`

exotations/dynamics_solvers/exotica_cartpole_dynamics_solver/src/cartpole_dynamics_solver.cpp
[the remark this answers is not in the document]
wxmerkt: We use many libraries that have similar naming for functionality ("Vector" or "vector" are just way too common of a name for objects in robotics!). Polluting namespaces can thus lead to clashes.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
include4eto: I was looking at the wrong thing..

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
include4eto: That was the original idea. However, on robots with a lot of joints this makes the XML a bit unreadable.

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    Eigen::MatrixXd Qx, Qu, Qxx, Quu, Qux, Quu_inv, Vxx;
+    Eigen::VectorXd Vx;
+
+    // V = prob_->GetStateCost(T - 1);
+    Vx = prob_->GetStateCostJacobian(T - 1);
+    Vxx = prob_->GetStateCostHessian(T - 1);
+
+    // concatenation axis for tensor products
+    Eigen::array<Eigen::IndexPair<int>, 1> dims = {Eigen::IndexPair<int>(1, 0)};
```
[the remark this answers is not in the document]
include4eto: Yes, but it's from `unsupported/Tensor`:
https://eigen.tuxfamily.org/dox-devel/unsupported/eigen_tensors.html#title14

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
include4eto: But one is `fx * Vxx` and the other - `fu * Vxx`? You could potentially use a temp for the second one..
Also, does't look like Croccodyl is doing this (at least in python):
https://gepgitlab.laas.fr/loco-3d/crocoddyl/blob/master/crocoddyl/ddp.py

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        // Q = prob_->GetStateCost(t) + prob_->GetControlCost(t) + V; // l + v
+        Qx = prob_->GetStateCostJacobian(t) + fx.transpose() * Vx;  // lx + fx @ Vx
+        Qu = prob_->GetControlCostJacobian(t) + fu.transpose() * Vx;
+        Qxx = prob_->GetStateCostHessian(t) + fx.transpose() * Vxx * fx;
+        Quu = prob_->GetControlCostHessian() + fu.transpose() * Vxx * fu;
+
+        // NOTE: Qux = Qxu for all robotics systems I have seen
+        //  this might need to be changed later on
+        // WARNING_NAMED("ControlLimitedDDPSolver", "Assuming fux = fxu. Make sure this is valid for your robot!");
```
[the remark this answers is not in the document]
include4eto: It spams like crazy.. Probably no point in having it there

exotations/solvers/exotica_ddp_solver/src/control_limited_ddp_solver.cpp
include4eto: It's clamping. It's not meant to go as far as the `double` limit.

exotica_core/include/exotica_core/tools/box_qp.h
include4eto: none

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,217 @@
...
+            prob_->Update(new_U.col(t), t);
+        prob_->SetCostEvolution(iteration, current_cost);
+    }
+
+    // store the best solution found over all iterations
+    for (int t = 0; t < T - 1; ++t)
+    {
+        solution.row(t) = global_best_U.col(t);
+        prob_->Update(global_best_U.col(t), t);
+    }
```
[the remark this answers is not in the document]
include4eto: That's done below.

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
include4eto: Did that.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
wxmerkt: When you get the initializer: If size == 1, set the same limit for all joints (i.e. setConstant() for the correct size in Eigen). If size == N, set as in the initializer. E.g. have a look at how joint velocity limits in the TimeIndexedProblems can be either 1 or N dimensions.

exotations/dynamics_solvers/exotica_quadrotor_dynamics_solver/src/quadrotor_dynamics_solver.cpp
include4eto: Now throwing exceptions on `fx` and `fu`, which are not implemented. The second-order derivatives are 0 by default.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,219 @@
...
+    ///\brief Computes the control gains for a the trajectory in the associated
+    ///     DynamicTimeIndexedProblem.
+    virtual void BackwardPass() = 0;
+
+    ///\brief Forward simulates the dynamics using the gains computed in the
+    ///     last BackwardPass;
+    /// @param alpha The learning rate.
+    /// @param ref_trajectory The reference state trajectory.
+    /// @return The cost associated with the new control and state trajectory.
+    double ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory);
```
wxmerkt: `const double` - it may also be a `const` since it only modifies the problem, not the solver. `alpha` isn't the learning rate but the line-search parameter (but yeah, it's kind of the same thing).

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,219 @@
...
+};
+
+template <typename Initializer>
+void AbstractDDPSolver<Initializer>::Solve(Eigen::MatrixXd& solution)
+{
+    if (!prob_) ThrowNamed("Solver has not been initialized!");
+    Timer planning_timer, backward_pass_timer, line_search_timer;
+
+    const int T = prob_->get_T();
+    const int NX = prob_->get_num_positions();
```
wxmerkt: Hmm, usually NX is position+velocity as NQ (which we dont use) would be position-only.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,219 @@
...
+
+        line_search_timer.Reset();
+        // forward pass to compute new control trajectory
+        // TODO: Configure line-search space from xml
+        // TODO (Wolf): What you are doing is a forward line-search when we may try a
+        //    backtracking line search for a performance improvement later - this just as an aside.
+        const Eigen::VectorXd alpha_space = Eigen::VectorXd::LinSpaced(10, 0.1, 1.0);
+        double current_cost = 0, best_alpha = 0;
+
+        Eigen::MatrixXd ref_trajectory = prob_->get_X();
```
wxmerkt: `const`

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,219 @@
...
+        {
+            double alpha = alpha_space(ai);
+            double cost = ForwardPass(alpha, ref_trajectory);
+
+            if (ai == 0 || cost < current_cost)
+            {
+                current_cost = cost;
+                new_U = prob_->get_U();
+                best_alpha = alpha;
+            }
```
wxmerkt: I think you can break this loop early if the cost is increasing.

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -0,0 +1,219 @@
...
+}
+
+template <typename Initializer>
+double AbstractDDPSolver<Initializer>::ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory)
+{
+    double cost = 0;
+    const int T = prob_->get_T();
+    const Eigen::VectorXd control_limits = dynamics_solver_->get_control_limits();
+
+    DynamicsSolverPtr dynamics_solver_ = prob_->GetScene()->GetDynamicsSolver();
```
wxmerkt: Why are you getting the pointer again when you have it in `dynamics_solver_`?

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/analytic_ddp_solver.h
```diff
@@ -0,0 +1,52 @@
...
+// \brief DDP solver using the analytic expansion of the value-function Q.
+//  Mayne, D. Q. (1966). "A second-order gradient method of optimizing non-linear discrete time systems".
+class AnalyticDDPSolver : public AbstractDDPSolver<AnalyticDDPSolverInitializer>, public Instantiable<AnalyticDDPSolverInitializer>
+{
+private:
+    ///\brief Computes the control gains for a the trajectory in the associated
+    ///     DynamicTimeIndexedProblem.
+    void BackwardPass() override;
+
+    void Instantiate(const AnalyticDDPSolverInitializer& init) override;
```
wxmerkt: Shouldn't this be public?

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/control_limited_ddp_solver.h
```diff
@@ -0,0 +1,53 @@
...
+// Control-limited DDP solver
+//  Control-limited Differential Dynamic Programming (Tassa, Mansard, Todorov, 2014)
+class ControlLimitedDDPSolver : public AbstractDDPSolver<ControlLimitedDDPSolverInitializer>, public Instantiable<ControlLimitedDDPSolverInitializer>
+{
+private:
+    ///\brief Computes the control gains for a the trajectory in the associated
+    ///     DynamicTimeIndexedProblem.
+    void BackwardPass() override;
+
+    void Instantiate(const ControlLimitedDDPSolverInitializer& init) override;
```
wxmerkt: public

exotations/solvers/exotica_ddp_solver/src/analytic_ddp_solver.cpp
```diff
@@ -0,0 +1,120 @@
...
+    parameters_ = init;
+    base_parameters_ = parameters_;
+}
+
+void AnalyticDDPSolver::BackwardPass()
+{
+    constexpr double min_clamp_ = -1e10;
+    constexpr double max_clamp_ = 1e10;
+    const int T = prob_->get_T();
+    const int NX = prob_->get_num_positions();
```
wxmerkt: As above, NX should be pos+vel

exotations/solvers/exotica_ddp_solver/src/control_limited_ddp_solver.cpp
```diff
@@ -0,0 +1,127 @@
...
+    const int T = prob_->get_T();
+    const int NX = prob_->get_num_positions();
```
wxmerkt: As above, NX != positions

exotica_core/include/exotica_core/tools/box_qp.h
```diff
@@ -0,0 +1,160 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+
+#ifndef EXOTICA_DDP_SOLVER_BOX_QP_H
```
wxmerkt: If in tools, change header-guard (and end with _)

exotica_core/include/exotica_core/tools/box_qp.h
```diff
@@ -0,0 +1,160 @@
...
+                           const Eigen::VectorXd& x_init)
+{
+    constexpr double epsilon = 1e-5;
+    constexpr double gamma = 0.1;
+    constexpr int max_iterations = 100;
+    return BoxQP(H, q, b_low, b_high, x_init, gamma, max_iterations, epsilon);
+}
+}  // namespace exotica
+
+#endif
```
wxmerkt: #endif  // name-of-header-guard

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
include4eto: See below

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -68,7 +68,7 @@ class AbstractDDPSolver : public MotionSolver
     /// @param alpha The learning rate.
     /// @param ref_trajectory The reference state trajectory.
     /// @return The cost associated with the new control and state trajectory.
-    double ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory);
+    double ForwardPass(const double alpha, const Eigen::MatrixXdRef ref_x, const Eigen::MatrixXdRef ref_u);
```
wxmerkt: We have Eigen::MatrixXdRefConst in exotica

exotations/solvers/exotica_ddp_solver/include/exotica_ddp_solver/abstract_ddp_solver.h
```diff
@@ -189,18 +191,17 @@ void AbstractDDPSolver<Initializer>::SpecifyProblem(PlanningProblemPtr pointer)
 }
 
 template <typename Initializer>
-double AbstractDDPSolver<Initializer>::ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory)
+double AbstractDDPSolver<Initializer>::ForwardPass(const double alpha, const Eigen::MatrixXdRef ref_x, const Eigen::MatrixXdRef ref_u)
```
wxmerkt: MatrixXdRefConst as above

exotations/solvers/exotica_ilqr_solver/include/exotica_ilqr_solver/ilqr_solver.h
```diff
@@ -69,7 +69,7 @@ class ILQRSolver : public MotionSolver, public Instantiable<ILQRSolverInitialize
     /// @param alpha The learning rate.
     /// @param ref_trajectory The reference state trajectory.
     /// @return The cost associated with the new control and state trajectory.
-    double ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory);
+    double ForwardPass(const double alpha, const Eigen::MatrixXdRef ref_x, const Eigen::MatrixXdRef ref_u);
```
wxmerkt: MatrixXdRefConst

exotations/solvers/exotica_ilqr_solver/src/ilqr_solver.cpp
```diff
@@ -99,19 +99,18 @@ void ILQRSolver::BackwardPass()
     }
 }
 
-double ILQRSolver::ForwardPass(double alpha, Eigen::MatrixXdRef ref_trajectory)
+double ILQRSolver::ForwardPass(const double alpha, const Eigen::MatrixXdRef ref_x, const Eigen::MatrixXdRef ref_u)
```
wxmerkt: MatrixXdRefConst

PR #599, opened 2019-06-13
separate vector and scalar task cost

christian-rauch:
This splits the task costs (per task) into its vector and scalar part. I.e. `GetTaskCost` provides the cost vector and `GetScalarTaskCost` provides its sum-of-squares as before.

### DIFF
--- a/exotica_core/src/problems/unconstrained_end_pose_problem.cpp
+++ b/exotica_core/src/problems/unconstrained_end_pose_problem.cpp
@@ -93,21 +93,9 @@ Eigen::RowVectorXd UnconstrainedEndPoseProblem::GetScalarJacobian() const
     return cost.jacobian.transpose() * cost.S * cost.ydiff * 2.0;
 }
 
-Eigen::VectorXd UnconstrainedEndPoseProblem::GetTaskError(const std::string& task_name) const
-{
-    for (int i = 0; i < cost.indexing.size(); ++i)
-    {
-        if (cost.tasks[i]->GetObjectName() == task_name)
-        {
-            return cost.ydiff.segment(cost.indexing[i].start, cost.indexing[i].length);
-        }
-    }
-    ThrowPretty("Cannot get scalar task cost. Task map '" << task_name << "' does not exist.");
-}
-
 double UnconstrainedEndPoseProblem::GetScalarTaskCost(const std::string& task_name) const
 {
-    const Eigen::VectorXd ydiff = GetTaskError(task_name);
+    const Eigen::VectorXd ydiff = cost.GetTaskError(task_name);
     return ydiff.transpose() * GetRho(task_name) * ydiff;
 }
 
--- a/exotica_core/src/problems/unconstrained_end_pose_problem.cpp
+++ b/exotica_core/src/problems/unconstrained_end_pose_problem.cpp
@@ -93,16 +93,22 @@ Eigen::RowVectorXd UnconstrainedEndPoseProblem::GetScalarJacobian() const
     return cost.jacobian.transpose() * cost.S * cost.ydiff * 2.0;
 }
 
-double UnconstrainedEndPoseProblem::GetScalarTaskCost(const std::string& task_name) const
+Eigen::VectorXd UnconstrainedEndPoseProblem::GetTaskError(const std::string& task_name) const
 {
     for (int i = 0; i < cost.indexing.size(); ++i)
     {
         if (cost.tasks[i]->GetObjectName() == task_name)
         {
-            return cost.ydiff.segment(cost.indexing[i].start, cost.indexing[i].length).transpose() * cost.rho(cost.indexing[i].id) * cost.ydiff.segment(cost.indexing[i].start, cost.indexing[i].length);
+            return cost.ydiff.segment(cost.indexing[i].start, cost.indexing[i].length);
         }
     }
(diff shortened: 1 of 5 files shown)


exotica_core/src/problems/unconstrained_end_pose_problem.cpp
```diff
@@ -93,18 +93,24 @@ Eigen::RowVectorXd UnconstrainedEndPoseProblem::GetScalarJacobian() const
     return cost.jacobian.transpose() * cost.S * cost.ydiff * 2.0;
 }
 
-double UnconstrainedEndPoseProblem::GetScalarTaskCost(const std::string& task_name) const
+Eigen::VectorXd UnconstrainedEndPoseProblem::GetTaskCost(const std::string& task_name) const
```
wxmerkt: Wouldn't `GetTaskError` or similar not be a more accurate name?

christian-rauch: I kept costs since it's the internally used name. Do you think it's better to have `GetTaskError` and `GetScalarTaskCost`?

wxmerkt: You are returning `ydiff` (y*-Phi) which is the TaskError. The cost is ydiff^T * S * ydiff. I would say TaskError here but not change GetScalarCost. Regarding GetTaskCost, I do not think anyone apart from you is using it at the moment? I let @VladimirIvan weigh in as well.

christian-rauch: Ah yeah, the `EndPoseTask` was called `cost` which contains the `ydiff`. That's why I stayed with `costs`. I am happy to rename if we agree on something.

christian-rauch: I just saw that `AICOSolver` and `BayesianIKSolver` have a `GetTaskCosts` method. They both return a scalar. Should they maybe be renamed to `GetScalarTaskCost`?

Maybe we should also add `GetScalarTaskCost` as pure virtual to the base class?

wxmerkt: The EndPoseTask should continue to be called cost. It is after all what it's used for and we should not rename it.

However, your new method wants to return the `ydiff`, i.e., _error_ of the cost. So `GetCostTaskError` could work. The `ScalarCost` is `ydiff^T * S * ydiff` (weighted sum of squares). 

AICO and BayesianIK should be untouched. Their GetTaskCosts are internal/private and used to update the messages. I have a new version of AICO/BayesianIK coming maybe in autumn where these things get revamped.

christian-rauch: Right, `GetTaskCosts` is private. I overlooked that.

Changed name to `GetTaskError` and added python binding.

wxmerkt: Thanks - I liked your previous comment so can we go for `GetCostTaskError`?

christian-rauch: Yes, I force pushed with the new method name `GetTaskError`.

exotica_core/src/problems/unconstrained_end_pose_problem.cpp
```diff
@@ -93,18 +93,24 @@ Eigen::RowVectorXd UnconstrainedEndPoseProblem::GetScalarJacobian() const
     return cost.jacobian.transpose() * cost.S * cost.ydiff * 2.0;
 }
 
-double UnconstrainedEndPoseProblem::GetScalarTaskCost(const std::string& task_name) const
+Eigen::VectorXd UnconstrainedEndPoseProblem::GetTaskError(const std::string& task_name) const
```
wxmerkt: Vlad and I had more discussions earlier on what would be the best way going forward with this - and here is what we decided upon:

1. Please place the `GetTaskError` method into `EndPoseTask` in `tasks.h`.
2. The tasks (`cost`, `equality`, `inequality`) of the corresponding problems should already be exposed as public. As such you can get your intended behaviour by calling `problem.cost.GetTaskError("my_task")`.
3. Do not move the `GetScalarTaskCost` in `UnconstrainedEndPoseProblem`, we will get to that at a later stage.

Background is that for the kinematic `TimeIndexedProblem`s, I did a clean-up in the past moving `GetGoal`, `SetGoal` etc. down into the tasks and using forwarding rather than duplicating the correct indexing in every problem. Mid-term - but not now - we want to move more that is task-related into those rather than the problems. As such, we want to introduce new functionality that is task-related always at the correct place.

christian-rauch: Makes sense. I moved the method to the task.
There is also a lot of duplication in the derived problems. Most problems are derived directly from the `PlanningProblem`. E.g. the `EndPoseProblem` and `UnconstrainedEndPoseProblem` share many common attributes and methods (`SetGoal`, `GetScalarCost`, `GetRho`, ...). From my point of view, the `EndPoseProblem` is a specialisation of the `UnconstrainedEndPoseProblem` since it adds constraints. Maybe `EndPoseProblem` should be renamed to `ConstrainedEndPoseProblem` and be derived from the `UnconstrainedEndPoseProblem`.

PR #601, opened 2019-06-17
Updates to Manipulability task map.

cmower:
[task_maps]
* Changed formulation from phi(x)=manip(x) to phi(x)=low - manip(x) so the task map can be used in both the cost function and as an inequality constraint (with lower bound low here).
* New LowerBound term in initializer.
* Updated documentation.
* Code formatting.

[examples]

* Rho now needs to be positive.

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
@@ -47,12 +47,13 @@ namespace exotica
 /// \f]
 /// that is based on the shape of the velocity ellipsoid where \f$J(x)\f$ is the manipulator Jacobian matrix.. The task map is expressed by
 /// \f[
-///   Phi(x) := m^- - m(x)
+///   \phi(x) := -m(x).
 /// \f]
-/// where \f$m^-\f$ is a lower bound for \f$m(x)\f$. If the task map is being used in the cost function then \f$m^-\f$ should be set to zero.
+///
+/// To use the task map as an inequality constraint the lower bound for \f$\phi\f$ should be set as a goal and each element should be negative.
 ///
 /// Note that
-///   - the associated value for \f$\rho\f$ <b>must</b> be positivein order to maximize the manipulability, and
+///   - the associated value for \f$\rho\f$ <b>must</b> be negative in order to maximize the manipulability, and
 ///   - derivatives of \f$\Phi\f$ are computed using finite differences.
 ///
 /// Todo
@@ -65,9 +66,8 @@ public:
     int TaskSpaceDim() override;
 
 private:
-    int n_end_effs_;               ///< Number of end-effectors.
-    int n_rows_of_jac_;            ///< Number of rows from the top to extract from full jacobian. Is either 3 (position) or 6 (position and rotation).
-    Eigen::VectorXd lower_bound_;  ///< When task map is a constraint, lower_bound_ is a lower bound for the inequality constraint, i.e. \f$\phi\geqm^-\f$ where \f$m^-\f$ is the lower bound.
+    int n_end_effs_;     ///< Number of end-effectors.
+    int n_rows_of_jac_;  ///< Number of rows from the top to extract from full jacobian. Is either 3 (position) or 6 (position and rotation).
 };
 }  // namespace exotica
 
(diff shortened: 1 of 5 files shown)


wxmerkt:
Can't you use "Goal" to set the lower bound?

exotations/exotica_core_task_maps/src/manipulability.cpp
```diff
@@ -33,23 +33,40 @@ REGISTER_TASKMAP_TYPE("Manipulability", exotica::Manipulability);
...
-
 void Manipulability::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi)
 {
     if (phi.rows() != TaskSpaceDim()) ThrowNamed("Wrong size of phi!");
 
     for (int i = 0; i < n_end_effs_; ++i)
     {
         const Eigen::MatrixXd& J = kinematics[0].jacobian[i].data.topRows(n_rows_of_jac_);
-        phi(i) = std::sqrt((J * J.transpose()).determinant());
+        phi(i) = lower_bound_(i) - std::sqrt((J * J.transpose()).determinant());
```
VladimirIvan: This can be done already using `SetGoal()`.

exotations/exotica_core_task_maps/include/exotica_core_task_maps/manipulability.h
```diff
@@ -41,31 +41,33 @@ namespace exotica
...
-/// where $J(x)$ is the manipulator Jacobian matrix.
+/// that is based on the shape of the velocity ellipsoid where \f$J(x)\f$ is the manipulator Jacobian matrix.. The task map is expressed by
+/// \f[
+///   Phi(x) := m^- - m(x)
+/// \f]
+/// where \f$m^-\f$ is a lower bound for \f$m(x)\f$. If the task map is being used in the cost function then \f$m^-\f$ should be set to zero.
 ///
 /// Note that
-///   - the associated value for \f$\rho\f$ <b>must</b> be negative in order to maximize the manipulability, and
+///   - the associated value for \f$\rho\f$ <b>must</b> be positivein order to maximize the manipulability, and
```
VladimirIvan: missing space

exotations/exotica_core_task_maps/src/manipulability.cpp
```diff
@@ -33,23 +33,22 @@ REGISTER_TASKMAP_TYPE("Manipulability", exotica::Manipulability);
...
-
 void Manipulability::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi)
 {
     if (phi.rows() != TaskSpaceDim()) ThrowNamed("Wrong size of phi!");
 
     for (int i = 0; i < n_end_effs_; ++i)
     {
         const Eigen::MatrixXd& J = kinematics[0].jacobian[i].data.topRows(n_rows_of_jac_);
-        phi(i) = std::sqrt((J * J.transpose()).determinant());
+        phi(i) = -std::sqrt((J * J.transpose()).determinant());
```
wxmerkt: This does not correspond to the description in `exotica_plugins` above

wxmerkt: I.e., you changed the sign here - but say above (in newly added comment) that the Rho needs to be negative to maximise. Is the minus required for the inequality to work?

cmower: I have updated `exotica_plugins` to be a little more descriptive.

Let `m = sqrt(det(J' * J))`. Above, `Goal` was requested to be used as the lower bound when the task map is used as an inequality constraint so `Phi` needs to take the form `Phi = -m` in order to satisfy both being set as a cost term and inequality. 
As a cost term, the objective term is `Rho * ydiff' * ydiff` where `ydiff = Phi - Goal = -m - Zeros`. To maximize the manipulability here `Rho` must be negative and the negative sign is squared so doesn't affect outcome.
As an inequality, the goal is `m >= LB <=> -m + LB <= 0` where `LB>0` is some strictly positive lower bound. For inequalities exotica specifies `ydiff = Phi - Goal <= 0`. Thus, in-order to achieve the desired inequality whilst corresponding with the cost-version of the task map and not introducing the `lower_bound_` parameter, `Phi` must be `-m` and `Goal<0`must take strictly negative values.

cmower: What might make sense is to perform a check in the `Update` that ensures `Goal` is strictly negative or zero. I'm not sure that it's possible to access `Goal` from within the task map?

wxmerkt: You can't, since that is _task_ and not _task-map_ dependent.

cmower: Ah of course, cheers.

PR #603, opened 2019-06-19
New GetNumberOfIterations method in planning problem

cmower:
New method to get the number of iterations, rather than 
```python
niter = len(problem.get_cost_evolution()[0])
```
now, 
```python
niter = problem.get_number_of_iterations()
```
Not sure why but in `GetNumberOfIterations` implementation I tried to return `cost_evolution_.size()` but this gave incorrect values. An explanation would be interesting to know.

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/src/planning_problem.cpp
+++ b/exotica_core/src/planning_problem.cpp
@@ -296,7 +296,7 @@ std::pair<std::vector<double>, std::vector<double>> PlanningProblem::GetCostEvol
 
 int PlanningProblem::GetNumberOfIterations() const
 {
-    return std::get<0>(GetCostEvolution()).size();
+    return static_cast<int>(std::get<0>(GetCostEvolution())).size();
 }
 
 double PlanningProblem::GetCostEvolution(int index) const
--- a/exotica_core/src/planning_problem.cpp
+++ b/exotica_core/src/planning_problem.cpp
@@ -294,6 +294,11 @@ std::pair<std::vector<double>, std::vector<double>> PlanningProblem::GetCostEvol
     return ret;
 }
 
+int PlanningProblem::GetNumberOfIterations() const
+{
+    return std::get<0>(GetCostEvolution()).size();
+}
+
 double PlanningProblem::GetCostEvolution(int index) const
 {
     if (index > -1 && index < cost_evolution_.size())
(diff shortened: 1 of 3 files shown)


exotica_core/src/planning_problem.cpp
```diff
@@ -294,6 +294,11 @@ std::pair<std::vector<double>, std::vector<double>> PlanningProblem::GetCostEvol
     return ret;
 }
 
+int PlanningProblem::GetNumberOfIterations() const
+{
+    return std::get<0>(GetCostEvolution()).size();
```
wxmerkt: Size is size_t, you have an implicit cast to int

exotica_python/src/pyexotica.cpp
```diff
@@ -721,6 +721,7 @@ PYBIND11_MODULE(_pyexotica, module)
         .def("get_number_of_problem_updates", &PlanningProblem::GetNumberOfProblemUpdates)
         .def("reset_number_of_problem_updates", &PlanningProblem::ResetNumberOfProblemUpdates)
         .def("get_cost_evolution", (std::pair<std::vector<double>, std::vector<double>>(PlanningProblem::*)() const) & PlanningProblem::GetCostEvolution)
+        .def("get_num_iterations", &PlanningProblem::GetNumberOfIterations)
```
wxmerkt: Please dont abbreviate. This is different from `get_number_of_problem_updates`

exotica_core/src/planning_problem.cpp
cmower: So should this line be `std::get<0>(GetCostEvolution()).size_t();`? Sorry, I am not so versed with C++.

wxmerkt: `static_cast<int>(std::get<0>(GetCostEvolution()).size());`

exotica_core/src/planning_problem.cpp
```diff
@@ -294,6 +294,11 @@ std::pair<std::vector<double>, std::vector<double>> PlanningProblem::GetCostEvol
     return ret;
 }
 
+int PlanningProblem::GetNumberOfIterations() const
+{
+    return static_cast<int>(std::get<0>(GetCostEvolution())).size();
```
wxmerkt: No, not like this. The static cast needs to close after size...

cmower: Sorry, should have copy/paste..resolved

Issue #573, opened 2019-05-18
Joint backward difference smoothing task maps broken


wxmerkt:
There must be a bug - Vlad reported the same thing yesterday - here with a few plots:

### No smoothing
![20190518_velocity_without_joint_smoothing_VELOCITY](https://user-images.githubusercontent.com/1664508/57972056-934a9900-798d-11e9-8b1b-76d2f3082596.png)

### Velocity smoothing
![20190518_velocity_with_joint_smoothing_VELOCITY](https://user-images.githubusercontent.com/1664508/57972058-96458980-798d-11e9-8b24-894e64dbeaac.png)

### Acceleration smoothing
![20190518_velocity_with_joint_smoothing_ACCEL](https://user-images.githubusercontent.com/1664508/57972061-9d6c9780-798d-11e9-87f6-e38271f07b58.png)

### Jerk smoothing
![20190518_velocity_with_joint_smoothing_JERK](https://user-images.githubusercontent.com/1664508/57972063-a198b500-798d-11e9-9f9a-d31887f47787.png)

wxmerkt:
A clearer picture:

### No smoothing
![0_no_smoothing](https://user-images.githubusercontent.com/1664508/57972366-2df8a700-7991-11e9-90f8-76de5aa3e258.png)

### Velocity smoothing
![1_velocity](https://user-images.githubusercontent.com/1664508/57972367-318c2e00-7991-11e9-841e-87eb60af5c4e.png)

### Acceleration smoothing
![2_acceleration](https://user-images.githubusercontent.com/1664508/57972368-3650e200-7991-11e9-951a-ad9b78465e4d.png)

### Jerk smoothing
![3_jerk](https://user-images.githubusercontent.com/1664508/57972370-394bd280-7991-11e9-9ad6-80ef8547dbcf.png)

### Velocity, acceleration, and jerk smoothing
![4_all](https://user-images.githubusercontent.com/1664508/57972374-3e108680-7991-11e9-8f2e-051c5485df8e.png)

wxmerkt:
@cmower Can you please have a look at this and comment on what may be going on?

wxmerkt:
@cmower Do you have any insights into what make things blow up with the backward-differencing smoothing maps?

PR #607, opened 2019-06-25
Updates to smoothing task maps and corresponding example.

cmower:
* Addressing issue #573. Main difference to note is that I have deprecated the dependency on `dt`. E.g., since `phi = (q + qb)/dt` the `1/dt` is only a scaling factor.
* Other changes that I have made are more to do with maintaining standards. 
* I have checked over the derivation for each smoothing task map and cannot find any errors. I think the issue you are seeing in #573 stems from the choice of `Rho`. This was one of the reasons I wrote the `interactive_cost_tuning` script. 

<!--
# Checklist

- add unit test(s)

--!>


wxmerkt:
Vlad, I was looking to get your feedback on whether removing the `dt` scaling makes sense...

PR #610, opened 2019-06-26
New AttachObjectLocal overloaded

cmower:
* New overloaded function accepts `Eigen::VectorXd` rather than `KDL::Frame`.
* Use of this is as follows

Before
```python
new_trans = np.array([0, 0, 1])
scene.attach_object_local('MyFrame', '', exo.KDLFrame(new_trans))
```
After
```python
new_trans = np.array([0, 0, 1])
scene.attach_object_local('MyFrame', '', new_trans)
```

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/src/scene.cpp
+++ b/exotica_core/src/scene.cpp
@@ -763,6 +763,11 @@ void Scene::AttachObjectLocal(const std::string& name, const std::string& parent
     attached_objects_[name] = AttachedObject(parent, pose);
 }
 
+void Scene::AttachObjectLocal(const std::string& name, const std::string& parent, const Eigen::VectorXd& pose)
+{
+    AttachObjectLocal(name, parent, GetFrame(pose));
+}
+
 void Scene::DetachObject(const std::string& name)
 {
     if (!HasAttachedObject(name)) ThrowPretty("'" << name << "' is not attached to the robot!");
--- a/exotica_examples/scripts/example_ik_look_at
+++ b/exotica_examples/scripts/example_ik_look_at
@@ -26,7 +26,7 @@ class Example(object):
 
     def update(self, event):
         # Set new look at target into world frame
-        self.problem.get_scene().attach_object_local('LookAtTarget', '', self.target_marker.position_exo)
+        self.problem.get_scene().attach_object_local('LookAtTarget', '', self.target_marker.position_exo.get_translation_and_quaternion())
 
         # Set start state
         self.problem.start_state = self.q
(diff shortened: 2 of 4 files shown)


exotica_core/src/scene.cpp
```diff
@@ -763,6 +763,37 @@ void Scene::AttachObjectLocal(const std::string& name, const std::string& parent
     attached_objects_[name] = AttachedObject(parent, pose);
 }
 
+void Scene::AttachObjectLocal(const std::string& name, const std::string& parent, const Eigen::VectorXd& pose)
+{
+    KDL::Frame frame;
```
wxmerkt: In Tools or Conversions there is already a function which does this for you

cmower:
@wxmerkt thanks for that! Resolved, squashed and force pushed.

wxmerkt:
I don't think this will compile as you haven't modified the pyexotica and it will have issues selecting.

cmower:
It does compile on my laptop. I have tried it on the `LookAt` example script, and works as expected.

wxmerkt:
It cannot - you are now having an overloaded function and do not specify which of them should be exposed to Python. And Travis confirms my suspicion with every single build failing. Cf. e.g. https://travis-ci.org/ipab-slmc/exotica/jobs/550926029#L1934

cmower:
You're correct. I don't understand what happened locally, perhaps I was running an old build and mistook it for the a newly built version that didn't exist. 

I have amended `pyexotica.cpp` with reference to [here](https://pybind11.readthedocs.io/en/master/classes.html#overloaded-methods). I modified under the assumption that we are using `-std=c++11`, from grepping the build files I couln't make out if we were using a C++ 14 compatible compiler or not so went with 11.

exotica_examples/scripts/example_ik_look_at
```diff
@@ -26,7 +26,7 @@ class Example(object):
 
     def update(self, event):
         # Set new look at target into world frame
-        self.problem.get_scene().attach_object_local('LookAtTarget', '', self.target_marker.position_exo)
+        self.problem.get_scene().attach_object_local('LookAtTarget', '', self.target_marker.position_exo.get_translation_and_quaternion())
```
wxmerkt: So now we are converting KDLFrame => Eigen => KDLFrame ... is this really cleaner?

cmower: In this example no, but often in scripts I find I am dealing with `np.array`s and it would be cleaner to pass in the `np.array` rather than `exo.KDLFrame(np.array([1,2,3]))` for example.

cmower: With this change I only wanted to demonstrate that it works with the least amount of modifications to `exotica_examples`.

PR #611, opened 2019-06-28
allow parsing empty vectors

christian-rauch:
This removes the exceptions when an empty vector (Eigen or STL) is parsed.

When a vector is used as an optional parameter, it often makes sense to have an empty default argument, e.g.:
```C++
Optional std::vector<std::string> Tasks = std::vector<std::string>();
```
or 
```C++
Optional Eigen::VectorXd Convergence = Eigen::VectorXd();
```
The parameter can then be ignored by testing the the vector insite the task map.

### DIFF
--- a/exotica_core/include/exotica_core/tools/conversions.h
+++ b/exotica_core/include/exotica_core/tools/conversions.h
@@ -39,6 +39,7 @@
 #include <vector>
 
 #include <exotica_core/tools/exception.h>
+#include <exotica_core/tools/printable.h>
 
 namespace Eigen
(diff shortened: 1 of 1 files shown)


wxmerkt:
I see the value in this. However, the original checks were due to different separators and people easily mixing them up, e.g., space and comma. So some failures would be due to erroneous input and lead to a silent failure. Perhaps add a warning rather than a throw?

christian-rauch:
I replaced the exception by `WARNING_NAMED`.

PR #616, opened 2019-06-30
Add generalised cost to DynamicTimeIndexedShootingProblem

### LINKED ISSUE #615: [DynamicTimeIndexedShootingProblem] Add general costs

wxmerkt:
cc: @include4eto 

This PR implements generalised costs for the dynamic problems and resolves #615.

It works reasonably well for LWR with generalised costs (depending on initialisation, of course). I also fixed some of your 'hacks' which introduced memory bugs - **always check user input from initializers prior to setting**.

**Unresolved bugs I have found:**
- The final state cost Qf, when set from Rho does not get picked up correctly. So I hacked the example by setting the Rho for T-2 and T-1, when in theory we should only need T-2. Perhaps the t for the linearisation of the cost is off?


exotica_core/include/exotica_core/problems/dynamic_time_indexed_shooting_problem.h
```diff
@@ -97,7 +106,19 @@ class DynamicTimeIndexedShootingProblem : public PlanningProblem, public Instant
     Eigen::VectorXd Dynamics(Eigen::VectorXdRefConst x, Eigen::VectorXdRefConst u);
     Eigen::VectorXd Simulate(Eigen::VectorXdRefConst x, Eigen::VectorXdRefConst u);
 
-private:
+protected:
+    /// \brief Checks the desired time index for bounds and supports -1 indexing.
+    inline void ValidateTimeIndex(int& t_in) const
```
include4eto: Good idea! this was getting annoying.

PR #628, opened 2019-07-11
[Dynamics] Various Fixes

include4eto:
Adds:
- Pinocchio symbolic derivatives and inverse dynamics.
- Default `fx/fu` implementation with finite differences.

Changes:
- Inverses in `iLQR, iLQG, AnalyticDDP, ControlLimitedDDP` now use a simple rule. Rate is configurable from `xml` configuration.
- Quadrotor forward dynamics using euler angles from quarternions.


exotica_core/include/exotica_core/dynamics_solver.h
```diff
@@ -133,6 +133,8 @@ class AbstractDynamicsSolver : public Object, Uncopyable, public virtual Instant
     Eigen::VectorXd get_control_limits();
     void set_control_limits(Eigen::VectorXd control_limits);
 
+    virtual ControlVector inverse_dynamics(const StateVector& state);
```
wxmerkt: InverseDynamics (unless it is a getter/setter, we use ThisMethod)

exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
```diff
@@ -272,6 +272,16 @@ void DynamicTimeIndexedShootingProblem::ReinitializeVariables()
     }
     cost.ReinitializeVariables(T_, shared_from_this(), cost_Phi);
 
+    if (this->parameters_.WarmStartWithInverseDynamics)
```
wxmerkt: Should this be in reinitialize or also in PreUpdate? What if someone changes X after setting the T (which calles Reinitialize)

exotica_core/include/exotica_core/dynamics_solver.h
```diff
@@ -150,7 +152,7 @@ class AbstractDynamicsSolver : public Object, Uncopyable, public virtual Instant
     Eigen::VectorXd control_limits_ = Eigen::VectorXd();  ///< ControlLimits. Default is empty vector.
 
     /// \brief Integrates the dynamic system from state x with controls u applied for one timestep dt using the selected integrator.
-    inline StateVector Integrate(const StateVector& x, const ControlVector& u);
+    virtual StateVector Integrate(const StateVector& x, const ControlVector& u);
```
wxmerkt: Why?

exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
include4eto: Duplicate it, you mean? I don't really see the use-case. Right now this is called in the constructor after everything else.

include4eto:
Accidentally committed the valkyrie demo with the last commit.. Oh, well.. it's in here now

exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
wxmerkt: Yes and that is the problem! If it is only in the constructor and not in PreUpdate, it will be ignored if you change `T` or anything else after construction...

exotations/dynamics_solvers/exotica_pinocchio_dynamics_solver/include/exotica_pinocchio_dynamics_solver/pinocchio_dynamics_solver.h
```diff
@@ -58,9 +59,11 @@ class PinocchioDynamicsSolver : public DynamicsSolver, public Instantiable<Pinoc
     StateVector f(const StateVector& x, const ControlVector& u) override;
     StateDerivative fx(const StateVector& x, const ControlVector& u) override;
     ControlDerivative fu(const StateVector& x, const ControlVector& u) override;
+    ControlVector InverseDynamics(const StateVector& x) override;
 
 private:
     pinocchio::Model model_;
+    std::unique_ptr<pinocchio::Data> pinocchio_data_;
```
wxmerkt: No need for a pointer I think

wxmerkt:
I am merging this as I require it to continue work. Please address the following in a follow-up:

(a) No need for the pointer
(b) Fix the inverse dynamics warm-start. It won't work if it's only set during the constructor. It needs to be done in the PreUpdate or ReinitializeVariables. Most likely PreUpdate since that is what _should_ be called as the first thing in Solve. If your solvers aren't doing that yet, time to add it ;-).

exotica_core/src/problems/dynamic_time_indexed_shooting_problem.cpp
include4eto: Ok, so why not put it just in `PreUpdate` then?

wxmerkt: Yep, please do - and check that your solvers are calling it at the beginning of Solve()

PR #637, opened 2019-07-20
OMPL Control RRT

include4eto:
- Adds an interface for ompl's control-rrt.
- Adds examples with dynamics and control-rrt.


exotations/solvers/exotica_control_rrt_solver/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(exotica_control_rrt_solver)
```
wxmerkt: Can we make this `exotica_ompl_control_solver` or similar, and then also expose KPIECE? Now that you've done the hard work to force OMPL to work... exposing an additional solver isn't that much more.

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
```diff
@@ -0,0 +1,135 @@
...
+    OMPLStatePropagator(
+        oc::SpaceInformationPtr si,
+        DynamicsSolverPtr dynamics_solver_) : oc::StatePropagator(si), space_(si), dynamics_solver_(dynamics_solver_) {}
+    void propagate(
+        const ob::State *state,
+        const oc::Control *control,
+        const double duration,
+        ob::State *result) const override
+    {
+        double t = timeStep_;
```
wxmerkt: Should you start with 0 here or with timeStep_?

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
```diff
@@ -0,0 +1,135 @@
...
+    void Integrate(ob::State *ob_x, const oc::Control *oc_u, double dt) const
+    {
+        const int NU = dynamics_solver_->get_num_controls();
+        const int NX = dynamics_solver_->get_num_positions() + dynamics_solver_->get_num_velocities();
+
+        double *x = ob_x->as<ob::RealVectorStateSpace::StateType>()->values;
+        double *u = oc_u->as<oc::RealVectorControlSpace::ControlType>()->values;
+
+        Eigen::VectorXd eig_x = Eigen::Map<Eigen::VectorXd>(x, NX);
+        Eigen::VectorXd eig_u = Eigen::Map<Eigen::VectorXd>(u, NU);
```
wxmerkt: Maybe add an assert (so will only work in debug) that the size of ob::State is the same as NX etc.)

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
```diff
@@ -0,0 +1,135 @@
...
+    ///\brief Binds the solver to a specific problem which must be pre-initalised
+    ///@param pointer Shared pointer to the motion planning problem
+    ///@return        Successful if the problem is a valid DynamicTimeIndexedProblem
+    void SpecifyProblem(PlanningProblemPtr pointer) override;
+
+private:
+    DynamicTimeIndexedShootingProblemPtr prob_;  ///!< Shared pointer to the planning problem.
+    DynamicsSolverPtr dynamics_solver_;          ///!< Shared pointer to the dynamics solver.
+
+    std::shared_ptr<oc::SimpleSetup> setup_;
```
wxmerkt: Can you use unique_ptr for all of these?

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
```diff
@@ -0,0 +1,190 @@
...
+    const int NX = prob_->get_num_positions() + prob_->get_num_velocities();
+    const double dt = dynamics_solver_->get_dt();
+    if (parameters_.Seed != -1) ompl::RNG::setSeed(parameters_.Seed);
+
+    std::shared_ptr<ob::RealVectorStateSpace> space(
+        std::make_shared<ob::RealVectorStateSpace>(NX));
+
+    state_bounds_ = std::make_shared<ob::RealVectorBounds>(NX);
+
+    for (int i = 0; i < NX; ++i)
```
wxmerkt: Check that `paramers_.StateLimits` has correct size, otherwise throw

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
```diff
@@ -0,0 +1,190 @@
...
+        std::vector<oc::Control *> controls = setup_->getSolutionPath().getControls();
+        std::vector<ob::State *> states = setup_->getSolutionPath().getStates();
+        std::vector<double> durations = setup_->getSolutionPath().getControlDurations();
+
+        Eigen::VectorXd sol_goal_state = Eigen::Map<Eigen::VectorXd>(
+            states.back()->as<ob::RealVectorStateSpace::StateType>()->values, NX);
+
+        T = 0;
+        for (int t = 0; t < controls.size(); ++t)
+            T += (int)(durations[t] / dt);
```
wxmerkt: C++ style cast

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
```diff
@@ -0,0 +1,190 @@
...
+                       !parameters_.ApproximateSolution))
+        {
+            if (debug_) HIGHLIGHT_NAMED("ControlRRTSolver", "No solution found.");
+            return;
+        }
+        else if (debug_)
+            HIGHLIGHT_NAMED("ControlRRTSolver", "Found solution.");
+
+        prob_->set_T(T);
+        prob_->PreUpdate();
```
wxmerkt: I see why you are doing this, but this would not work if someone did change something about the problem (Q, Xstar, taskmaps which dont work here anyways). There is no way to do this without changing the problem inside the solver?

exotica_examples/resources/configs/dynamic_time_indexed/13_control_limited_ddp_quadrotor.xml
```diff
@@ -1,13 +1,13 @@
 <?xml version="1.0" ?>
 <PlannerDemoConfig>
-    <ControlLimitedDDPSolver Name="ControlLimitedDDPSolver">
+    <ILQGSolver Name="ControlLimitedDDPSolver">
```
wxmerkt: Ummm, conflicts with name of the example

exotica_examples/scripts/example_dynamic_time_indexed
```diff
@@ -37,34 +39,45 @@ print('Running with config {0}'.format(configs[problem_idx]))
 solver = exo.Setup.load_solver(configs[problem_idx])
 problem = solver.get_problem()
 solution = solver.solve()
+if solution.shape[0] == 0:
```
wxmerkt: Maybe write an error/reason why it exits? ["Could not find a solution/did not converge"]

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
include4eto: What do you mean only work in debug?

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
include4eto: I don't see how, since you don't know the number of timesteps before RRT converges.

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
wxmerkt: If you use `assert`, it will only be executed when compiled in Debug. Assert statements are removed from release. So it's good to test something but without affecting performance in Release.

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
wxmerkt: Yes, but instead of changing the problem you can use the start-state and integrator from dynamics-solver to do the roll-out just in the motion solver. No need to change problem.

include4eto: But then the solution will have a different `T` from the problem. Which will break attempts to playback the solution by calling `update` on the problem for all timesteps.

wxmerkt: True, but that has to be handled in user code I feel.

exotations/solvers/exotica_control_rrt_solver/CMakeLists.txt
include4eto: Can we leave this a separate issue/ PR? It's actually quite a bit of change in the structure of the thing (esp. initializers & exposing the relevant parameters for RRT/KPIECE) and I think I'll leave it for after submission.

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
include4eto: I actually removed these. I was getting `double free or corruption (out)` and I added a bunch of them to be safe but I removed them and now it's fine.

It all comes from `SimpleSetup` taking a bunch of pointers in function args.

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
include4eto: Moved that check to `SpecifyProblem`.

exotica_examples/scripts/example_dynamic_time_indexed
include4eto: Added a print below.

exotations/solvers/exotica_control_rrt_solver/src/control_rrt_solver.cpp
include4eto: I'm not convinced - wouldn't you want the problem to contain a solution once you've called `solve` on it?

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
include4eto: added

include4eto:
@wxmerkt Added a bunch more fixes here, including:

- https://github.com/ipab-slmc/exotica/issues/640
- https://github.com/ipab-slmc/exotica/issues/641

exotations/solvers/exotica_control_rrt_solver/CMakeLists.txt
include4eto: Nevermind, did it.

However, i still had the `libraryunload` poco exception when having the implementation for the base class in a separate `.cpp` file. This is similar to `AbstractDDPSolver`.

@VladimirIvan The other OMPL solver seems to make this work. I don't know how or why?

include4eto: https://github.com/ipab-slmc/exotica/issues/644

exotations/solvers/exotica_ompl_control_solver/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(exotica_ompl_control_solver)
```
wxmerkt: solvers (plural), there are multiple

exotations/solvers/exotica_ompl_control_solver/package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>exotica_ompl_control_solver</name>
```
wxmerkt: plural, as above

exotica_core/include/exotica_core/dynamics_solver.h
```diff
@@ -130,14 +130,15 @@ class AbstractDynamicsSolver : public Object, Uncopyable, public virtual Instant
     void SetIntegrator(std::string integrator_in);
 
     /// \brief Returns the control limits vector.
-    Eigen::VectorXd get_control_limits();
-    void set_control_limits(Eigen::VectorXd control_limits);
+    Eigen::VectorXd get_control_limits_low();
```
wxmerkt: For get we usually return a matrix where the first column is the lower and the second column is the upper limit. Could you please change it to that as well? It matches the kinematic limits etc.

exotica_examples/resources/configs/dynamic_time_indexed/13_control_limited_ddp_quadrotor.xml
```diff
@@ -1,13 +1,13 @@
 <?xml version="1.0" ?>
 <PlannerDemoConfig>
-    <ControlLimitedDDPSolver Name="ControlLimitedDDPSolver">
+    <AnalyticDDPSolver Name="ControlLimitedDDPSolver">
```
wxmerkt: File name says ControlLimited

exotica_python/src/pyexotica.cpp
```diff
@@ -742,7 +742,8 @@ PYBIND11_MODULE(_pyexotica, module)
         .def("reset_number_of_problem_updates", &PlanningProblem::ResetNumberOfProblemUpdates)
         .def("get_cost_evolution", (std::pair<std::vector<double>, std::vector<double>>(PlanningProblem::*)() const) & PlanningProblem::GetCostEvolution)
         .def("get_number_of_iterations", &PlanningProblem::GetNumberOfIterations)
-        .def("is_valid", &PlanningProblem::IsValid);
+        .def("is_valid", &PlanningProblem::IsValid)
+        .def_readonly("termination_criterion", &PlanningProblem::termination_criterion);
```
wxmerkt: def_property_readonly

exotations/solvers/exotica_ompl_control_solver/CMakeLists.txt
include4eto: Yes, but it's `ompl_solver`.

exotations/solvers/exotica_ompl_control_solver/package.xml
include4eto: Again, it's `ompl_solver` that we currently have.

exotations/solvers/exotica_control_rrt_solver/include/exotica_control_rrt_solver/control_rrt_solver.h
include4eto: Actually, this doesn't work and I'm not sure it can. See
https://ompl.kavrakilab.org/classompl_1_1base_1_1RealVectorStateSpace_1_1StateType.html#details

I don't know how to cast from `RealVectorStateSpace::StateType` back to `RealVectorStateSpace` or if that's even possible..

exotations/solvers/exotica_ompl_control_solver/CMakeLists.txt
wxmerkt: ok

exotica_python/src/pyexotica.cpp
include4eto: But it's not a property? It's just a public member.

PR #646, opened 2019-08-01
require msgpack

christian-rauch:
Small fix to make CMake fail if msgpack is not found.


christian-rauch:
Do you mean the wrong `depend` tag for ZeroMQ (that makes the kinetic CI fail)? I don't see any other warnings during my builds.

christian-rauch:
The pkg-config file `msgpack.pc` is only available since bionic/melodic: https://packages.ubuntu.com/search?searchon=contents&keywords=msgpack.pc&suite=bionic. There is no pkg-config for xenial. We need a custom CMake module.

VladimirIvan:
The msg-pack version on xenial is too low. It doesn't support the features we need. The whole meshcat visualizer is disabled in that case. The cmake file there would not serve any purpose.

exotica_core/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 find_package(TinyXML2 REQUIRED)
 find_package(ZeroMQ REQUIRED)
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(MSGPACK msgpack)
+pkg_check_modules(MSGPACK REQUIRED msgpack)
```
wxmerkt: We could use the `MSGPACK_FOUND` to append the requirement to the list that is set in catkin_package.

exotica_core/package.xml
```diff
@@ -28,7 +28,7 @@
   <depend>eigen_conversions</depend>
   <depend>tf_conversions</depend>
   <depend>tinyxml2</depend>
-  <depend>libzmq3-dev</depend>
+  <depend>zeromq3</depend>
```
wxmerkt: Is this the correct key or does it change behaviour?

wxmerkt: I just saw your comment:
> Edit: There are two rosdep keys for ZeroMQ: libzmq3-dev and zeromq3. The package actually got installed.

So, can we revert to the other one?

christian-rauch: I opened an issue at rosdistro (https://github.com/ros/rosdistro/issues/21879) and wanted to get feedback on this. IMHO, `zeromq3` is the more generic name of the library because `libzmq3-dev` is already the name of the Debian/Ubuntu package. In any case, they both result in installing the package `libzmq3-dev` on Debian/Ubuntu. I was just hoping that this issue gets addressed in rosdistro before merging this.

exotica_core/CMakeLists.txt
christian-rauch: I think we could also disable the meshcat visualiser if msgpack is not installed and print an informative message that an optional feature requires msgpack.

exotica_core/package.xml
wxmerkt: I am happy to stay with the more generic one then. Thanks for following this up upstream

christian-rauch:
@VladimirIvan From https://github.com/ipab-slmc/exotica/blob/e2c59d80d24fa49679670448e2585e3246229e5a/exotica_core/include/exotica_core/visualization_meshcat_types.h#L44-L51 it looks like there is a workaround for Ubuntu 16.04 with msgpack < v1.
Shall we disable the meshcat visualiser for msgpack versions < 2 or keep this workaround?

christian-rauch:
I went ahead and made the `VisualizationMeshcat` optional. A warning is shown if MessagePack is not installed and `VisualizationMeshcat` gets disabled.

wxmerkt:
> A warning is shown if MessagePack is not installed and VisualizationMeshcat gets disabled.

This will result in constant "unstable" emails from the ROS buildfarm as it is 'stable' on Melodic and 'with warning' on Kinetic. And thus 'unstable' between versions.

christian-rauch:
Correct. I changed it to STATUS.

exotica_core/CMakeLists.txt
```diff
@@ -23,7 +23,13 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 find_package(TinyXML2 REQUIRED)
 find_package(ZeroMQ REQUIRED)
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(MSGPACK msgpack)
+pkg_check_modules(MSGPACK QUIET msgpack)
```
wxmerkt: Did we actually test and verify that Meshcat is broken on Kinetic or are we just deactivating it because the zeromq didnt have the pkg-config?

christian-rauch: I interpreted Vlad's comment as if the version in Ubuntu 16.04 is incompatible with the codebase.

VladimirIvan: I compiled it on 16.04. It compiles with warning but the features are not available in msgpack so I disabled it in the code but left it in CMAKE so that we don't have to propagate that logic to exotica_python and the examples. They will just fail gracefully when run.

Issue #648, opened 2019-08-06
Erroneous output for transformation get functions


cmower:
I noticed the transform frames being outputted by `scene.fk('EffFrame', '').get_frame()` (where `EffFrame` is just a constant frame wrt `lwr_arm_7_link`) seems to be incorrect. See image below.

![potential_bug](https://user-images.githubusercontent.com/8171456/62544599-fdb9f980-b857-11e9-9385-c4ba728b8254.png)

I am printing to the terminal (bottom) two transform frames. The top (`EffFrame in world_frame (without inverse)`) is that outputted by `scene.fk('EffFrame', '').get_frame()`. Upon inspection, the rotation part is the inverse of what it should be (`EffFrame in world_frame (with inverse)` shows what the transform should be).

wxmerkt:
Can you provide a minimally working example?

cmower:
Do you mean a potential fix for the problem or an example, in code, of the issue? In the above example, to generate the top transform I (essentially) use 
```python
f1 = scene.fk('EffFrame', '').get_frame()
print f1
```
and in the bottom I use 
```python
f2 = scene.fk('EffFrame', '').get_frame()
f2[:3,:3] = f2[:3,:3].transpose()
print f2
```
The `f1` frame, as printed in the picture above, is the wrong transform frame.

cmower:
I looked into this a little deeper and I am certain it is a bug and can explain why. 

Eigen maps the data from the `data` array as such
```c++
int array[9];
for(int i = 0; i < 9; ++i) array[i] = i;
cout << Map<Matrix3i>(array) << endl;
```
and the output is 
```
 0  3  6
 1  4  7
 2  5  8
```
see documentation [here](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html).

Exotica extracts the rotation matrix from the KDL frame directly from the frames `data` attribute, see [here](https://github.com/ipab-slmc/exotica/blob/cad9fbad9906cefed16cf0d16fbe05dc9c163a23/exotica_core/src/tools/conversions.cpp#L96), and passes it to the Eigen `Map`.

However, KDL stores the rotation matrix as follows (elements are the corresponding indices for illustration).
```
data = {0, 1, 2, 3, 4, 5, 6, 7, 8}

    [ 0 1 2 ]
M = [ 3 4 5 ]
    [ 6 7 8 ]
```
[...]

PR #650, opened 2019-08-08
[conversions] bug fix in GetFrame

cmower:
Fix for #648 
<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/src/tools/conversions.cpp
+++ b/exotica_core/src/tools/conversions.cpp
@@ -90,7 +90,7 @@ KDL::Frame GetFrameFromMatrix(Eigen::MatrixXdRefConst val)
 
 Eigen::MatrixXd GetFrame(const KDL::Frame& val)
 {
-    Eigen::Affine3d ret;
+    Eigen::Isometry3d ret;
     tf::transformKDLToEigen(val, ret);
     return ret.matrix();
 }
--- a/exotica_core/src/tools/conversions.cpp
+++ b/exotica_core/src/tools/conversions.cpp
@@ -27,6 +27,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 //
 
+#include <eigen_conversions/eigen_kdl.h>
 #include <algorithm>
 
 #include <exotica_core/tools/conversions.h>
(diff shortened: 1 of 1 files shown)


exotica_core/src/tools/conversions.cpp
```diff
@@ -93,7 +93,7 @@ Eigen::MatrixXd GetFrame(const KDL::Frame& val)
     ret(0, 3) = val.p(0);
     ret(1, 3) = val.p(1);
     ret(2, 3) = val.p(2);
-    ret.block(0, 0, 3, 3) = Eigen::Map<const Eigen::Matrix3d>(val.M.data);
+    ret.block(0, 0, 3, 3) = Eigen::Map<const Eigen::Matrix3d>(val.M.data).transpose();
```
wxmerkt: Cf. https://github.com/ipab-slmc/exotica/issues/648#issuecomment-519654484

exotica_core/src/tools/conversions.cpp
```diff
@@ -89,21 +90,19 @@ KDL::Frame GetFrameFromMatrix(Eigen::MatrixXdRefConst val)
 
 Eigen::MatrixXd GetFrame(const KDL::Frame& val)
 {
-    Eigen::MatrixXd ret = Eigen::MatrixXd::Identity(4, 4);
-    ret(0, 3) = val.p(0);
-    ret(1, 3) = val.p(1);
-    ret(2, 3) = val.p(2);
-    ret.block(0, 0, 3, 3) = Eigen::Map<const Eigen::Matrix3d>(val.M.data).transpose();
-    return ret;
+    Eigen::Affine3d ret;
```
wxmerkt: Please use Isometry, not Affine.

cmower: Any reason for this?

wxmerkt: Yes. If you would do an inverse on the Affine, it runs SVD which is expensive. Rigid body motions do not do shearing, scaling etc. so no need to have. Isometry is much cheaper with the inverse as it's direct.

cmower:
Yes, all returning correct values. No problem!

PR #661, opened 2019-09-04
Expose GetNumControlledJoints to python scene

cmower:
Minor addition, exposes `GetNumControlledJoints` in `KinematicTree` to python scene. 

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotica_core/include/exotica_core/scene.h
+++ b/exotica_core/include/exotica_core/scene.h
@@ -87,6 +87,7 @@ public:
     std::string GetRootJointName();
 
     exotica::KinematicTree& GetKinematicTree();
+    int GetNumberOfControlledJoints() const;
     std::vector<std::string> GetControlledJointNames();
     std::vector<std::string> GetModelJointNames();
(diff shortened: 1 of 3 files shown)


wxmerkt:
This has already been exposed. You can get the KinematicTree from the scene via: `scene.get_kinematic_tree()` and then the number of controlled joints from there:

https://github.com/ipab-slmc/exotica/blob/9e1da605f2b191a17410b2f19be19b5e256c7abc/exotica_python/src/pyexotica.cpp#L1151

cmower:
Similar functions are implemented like `GetControlledLinkNames` and `GetModelJointNames` that are just by-passing `get_kinematic_tree`. This was only to bring `GetNumControlledJoints` in sync with these (I had a simple use-case for it and thought it would be a quick addition).

wxmerkt:
The other ones are for historic reasons and are meant to be removed. Is something preventing the correct use of the already exposed method?

cmower:
Ahh ok, I didn't realize this, no need for this then

PR #666, opened 2019-09-13
[exotica] Fix typo in documentation overview

the-raspberry-pi-guy:

### DIFF
--- a/exotica/doc/overview.rst
+++ b/exotica/doc/overview.rst
@@ -77,7 +77,7 @@ The system model provides an interface to answer kinematic queries. A query can 
  - Name of the tip frame (Frame A)
  - Offset of the tip frame
  - Name of the base frame (Frame B)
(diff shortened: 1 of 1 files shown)


exotica/doc/overview.rst
```diff
@@ -77,7 +77,7 @@ The system model provides an interface to answer kinematic queries. A query can
  - Name of the tip frame (Frame A)
  - Offset of the tip frame
  - Name of the base frame (Frame B)
- - Offset the of base frame
+ - Offset of the base frame
```
wxmerkt: While we are at it, let's make it semantically correct: `offset from` instead of `offset of`

wxmerkt:
@the-raspberry-pi-guy Thank you very much for spotting and fixing this :-). Can you please address the above comment in a new PR? This one got merged a bit too soon. Thank you!

PR #667, opened 2019-09-13
const reference getter for KinematicTree

christian-rauch:
return some `KinematicTree` members as const reference

### DIFF
--- a/exotica_core/include/exotica_core/kinematic_tree.h
+++ b/exotica_core/include/exotica_core/kinematic_tree.h
@@ -134,15 +134,15 @@ class KinematicTree : public Uncopyable
 {
 public:
     void Instantiate(std::string joint_group, robot_model::RobotModelPtr model, const std::string& name);
-    std::string GetRootFrameName() const;
-    std::string GetRootJointName() const;
+    const std::string& GetRootFrameName() const;
+    const std::string& GetRootJointName() const;
     robot_model::RobotModelPtr GetRobotModel() const;
     BaseType GetModelBaseType() const;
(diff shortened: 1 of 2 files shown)


exotica_core/include/exotica_core/kinematic_tree.h
```diff
@@ -134,15 +134,15 @@ class KinematicTree : public Uncopyable
...
+    const std::string& GetRootFrameName() const;
+    const std::string& GetRootJointName() const;
     robot_model::RobotModelPtr GetRobotModel() const;
     BaseType GetModelBaseType() const;
     BaseType GetControlledBaseType() const;
     std::shared_ptr<KinematicResponse> RequestFrames(const KinematicsRequest& request);
     void Update(Eigen::VectorXdRefConst x);
     void ResetJointLimits();
-    Eigen::MatrixXd GetJointLimits() const { return joint_limits_; }
+    const Eigen::MatrixXd& GetJointLimits() const { return joint_limits_; }
```
wxmerkt: I want to test this change. I had this previously with a const-reference-getter and it caused bugs with seg-faults with certain versions of Eigen and optimisation flags turned on (`-O3`). Took me a while to resolve and required 'odd' workarounds (loops instead of vector assignments etc).

christian-rauch: Do you have an idea how to test this reliably? I haven't encountered strange issues with this so far. After all, `joint_limits_` is just a classic class member and the getter should behave as if we directly access this member.

PR #670, opened 2019-09-26
Updated functionality of EffBox task map

cmower:
[task_maps] Rather than returning whole vector for lower/upper limits, now user specifies the limits of which end-effector they want to have access to
[examples] Updated example script to show how python interface to EffBox task map is used to get lower/upper limits

<!--
# Checklist

- add unit test(s)

--!>

### DIFF
--- a/exotations/exotica_core_task_maps/src/eff_box.cpp
+++ b/exotations/exotica_core_task_maps/src/eff_box.cpp
@@ -71,14 +71,16 @@ void EffBox::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::Ma
     if (debug_ && Server::IsRos()) PublishObjectsAsMarkerArray();
 }
 
-Eigen::VectorXd EffBox::GetLowerLimit() const
+Eigen::Vector3d EffBox::GetLowerLimit(const int eff_id) const
 {
-    return eff_lower_;
+    if (eff_id < 0 || eff_id >= n_effs_) ThrowNamed("Given eff_id (" << eff_id << ") is out of range [0, " << n_effs_ << ")!");
+    return eff_lower_.segment<3>(3 * eff_id);
 }
 
-Eigen::VectorXd EffBox::GetUpperLimit() const
+Eigen::Vector3d EffBox::GetUpperLimit(const int eff_id) const
 {
-    return eff_upper_;
+    if (eff_id < 0 || eff_id >= n_effs_) ThrowNamed("Given eff_id (" << eff_id << ") is out of range [0, " << n_effs_ << ")!");
+    return eff_upper_.segment<3>(3 * eff_id);
 }
 
 void EffBox::PublishObjectsAsMarkerArray()
--- a/exotations/exotica_core_task_maps/include/exotica_core_task_maps/eff_box.h
+++ b/exotations/exotica_core_task_maps/include/exotica_core_task_maps/eff_box.h
@@ -52,8 +52,8 @@ public:
 
     void PublishObjectsAsMarkerArray();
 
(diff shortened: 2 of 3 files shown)


cmower:
I'm not sure what this error is but all builds and works fine locally.

VladimirIvan:
Try to compile it in debug mode on your local machine as well: 
```
catkin config --cmake-args -DCMAKE_BUILD_TYPE=Debug
catkin build
```
It sometimes catches bugs that the release mode doesn't (e.g., Eigen static asserts). It looks like  travis may have just messed up that particular build and everything is OK.

exotations/exotica_core_task_maps/src/eff_box.cpp
```diff
@@ -71,14 +71,16 @@ void EffBox::Update(Eigen::VectorXdRefConst x, Eigen::VectorXdRef phi, Eigen::Ma
...
-    return eff_lower_;
+    if (eff_id < 0 || eff_id >= n_effs_) ThrowNamed("Given eff_id (" << eff_id << ") is out of range [0, " << n_effs_ << ")!");
+    return eff_lower_.segment<3>(3 * eff_id);
 }
 
-Eigen::VectorXd EffBox::GetUpperLimit() const
+Eigen::Vector3d EffBox::GetUpperLimit(const int eff_id) const
 {
-    return eff_upper_;
+    if (eff_id < 0 || eff_id >= n_effs_) ThrowNamed("Given eff_id (" << eff_id << ") is out of range [0, " << n_effs_ << ")!");
```
wxmerkt: I think you need to check against n_effs_ - 1.

cmower: Let there be two end-effectors that you are constraining using `EffBox`, then `n_effs_ = 2` and the only possible values for `eff_id` is `0` or `1`. 

If I give an `eff_id` from the set `{..., -3, -2, -1}` then 
```
eff_id < 0 || eff_id >= n_effs_ -> True
   True    or    False
```
If I give an `eff_id` from the set `{2, 3, 4, ...}` then
```
eff_id < 0 || eff_id >= n_effs_ -> True
   False   or    True      
```
If I give an `eff_id` from the set `{0, 1}` then 
```
eff_id < 0 || eff_id >= n_effs_ -> False
   False   or    False
```

cmower:
> Try to compile it in debug mode on your local machine as well:
> 
> ```
> catkin config --cmake-args -DCMAKE_BUILD_TYPE=Debug
> catkin build
> ```
> 
> It sometimes catches bugs that the release mode doesn't (e.g., Eigen static asserts). It looks like travis may have just messed up that particular build and everything is OK.

Tried this and no warnings/errors to report. Build succeeds as expected. In this case, I think you're correct that travis is having issues.

All good to merge then? It's quite a simple addition...

exotations/exotica_core_task_maps/src/eff_box.cpp
wxmerkt: No need for such a mathematical explanation ;-). My argument was `n_end_effs_` is the number of end-effectors. So when you have `2` (ids `0` and `1`), `n_end_effs_` is 2.

Nevermind, realised I didn't correct interpret the `>=` in this case.

wxmerkt:
When Travis fails, please have a look at the logs there on the source of failure. In this case it says time-out (>50mins) for both Debug configurations. That is not a good sign. You can restart individual sub-builds without the whole build when you login to Travis using your GitHub account (I just did).

If it times out again, this PR (albeit a minor change) may have pushed it over in terms of compilation time or memory exhaustion. This wouldn't be a good sign as it would reduce our overall CI coverage and we may have to either (a) ask travis for runtime extension or (b) set up new CI infrastructure.

PR #675, opened 2019-10-23
Pinocchio as third-party library

wxmerkt:
This PR testst whether we can 

(a) upgrade Pinocchio to the latest 2.1.9 (and work around a bug with boost::variant)
(b) use a third-party release mechanism

This will require re-releasing eigenpy and pinocchio via the ROS buildfarm (breaking change). It will make our development a lot easier though and also fix the buildsystem problems we have been seeing with Crocoddyl.

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,8 +1,8 @@
-- git:
-    uri: https://github.com/stack-of-tasks/eigenpy.git
-    version: devel
-    local-name: eigenpy
-- git:
-    uri: https://github.com/wxmerkt/pinocchio.git
-    version: topic/wxm-add-package.xml
-    local-name: pinocchio
+#- git:
+#    uri: https://github.com/stack-of-tasks/eigenpy.git
+#    version: devel
+#    local-name: eigenpy
+#- git:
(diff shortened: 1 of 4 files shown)


.travis.rosinstall
```diff
@@ -1,6 +1,8 @@
 #- git:
-#    uri: https://github.com/ipab-slmc/eigenpy_catkin.git
+#    uri: https://github.com/stack-of-tasks/eigenpy.git
+#    version: devel
 #    local-name: eigenpy
 #- git:
-#    uri: https://github.com/ipab-slmc/pinocchio_catkin.git
+#    uri: https://github.com/stack-of-tasks/pinocchio.git
+#    version: devel
```
VladimirIvan: Can you make these point to a commit hash instead of `devel`?
Devel tends to get broken at 11pm on Fridays.

wxmerkt: This is not being used right now (commented). When I activate it temporarily on the buildserver, I will fix it to the corresponding version. We are directly using the ROS buildfarm binaries (not yet synced, but sync is upcoming)

PR #690, opened 2020-02-24
[exotica_python] Add MoveIt Scene to SDF converter

wxmerkt:
@rshi159 This is the basic MoveIt Scene to SDF conversion script. It currently only has box support - please take it from here and expand it / clean it up as required and make a PR :-)

cc: FYI @mfinean


VladimirIvan:
It's not that I'm a huge fan of `PyKDL` but what is reasoning behind including `transformations.py`?

wxmerkt:
This is code from ~2 years ago meant as a starter to evolve from (it only supports boxes so far). I agree we have pyexotica and PyKDL transformation packages and this should be replaced. This is to bootstrap the work @rshi159 is doing :-).

PR #691, opened 2020-02-26
Kinematic element visual name duplicate fix

VladimirIvan:
- Uniquely names visual elements with duplicate names by appending an index: `Shelf` will become `Shelf_0`, `Shelf_1`, ...

This bug was due to the scene loading reusing the object name for all visual elements.
Meshacat visualiser would not display the scene objects correctly because it uses the unique name as an index. Duplicates would cause it to overwrite the object with the same name.

### DIFF
--- a/exotica_core/src/scene.cpp
+++ b/exotica_core/src/scene.cpp
@@ -619,7 +619,7 @@ void Scene::UpdateSceneFrames()
                 if (it != visual_map.end())
                 {
                     it->second++;
-                    name = name + std::to_string(it->second);
+                    name = name + "_" + std::to_string(it->second);
                 }
                 else
                 {
--- a/exotica_core/src/scene.cpp
+++ b/exotica_core/src/scene.cpp
@@ -595,6 +595,7 @@ void Scene::UpdateSceneFrames()
     kinematica_.ResetModel();
 
     // Add world objects
+    std::map<std::string, int> visual_map;
     for (const auto& object : *ps_->getWorld())
     {
         if (object.second->shapes_.size())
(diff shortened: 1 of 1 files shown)


exotica_core/src/scene.cpp
```diff
@@ -612,7 +613,19 @@ void Scene::UpdateSceneFrames()
...
                 Eigen::Isometry3d trans = obj_transform.inverse() * shape_transform;
                 VisualElement visual;
-                visual.name = object.first;
+                std::string name = object.first;
+                // Check for name duplicates after loading from scene files
+                const auto& it = visual_map.find(name);
+                if (it != visual_map.end())
+                {
+                    it->second++;
+                    name = name + std::to_string(it->second);
```
wxmerkt: This makes it `Shape1` not `Shape_1` as in the PR description. Possibly worth printing a warning as well.

VladimirIvan: Fixed.
Adding warnings will print them every time objects get added/removed. It would print warnings for calls that didn't cause the issue. I'd rather keep it without a warning.

wxmerkt: Oh, now I understand - so this is equivalent to `Shelf_collision_{n}`? Then perhaps add `_visual_` as a substring for consistency?

VladimirIvan: The collision shapes do all get `_collision_#` prefix. For visuals, only the ones that got duplicated names (loaded from a scene file with multiple shapes defined). The naming is: `Shelf`, `Shelf_0`, `Shelf_1` ... when there are multiple meshes and correctly `Shelf` when there is only one. I'd rather keep it simple for now.

wxmerkt: Ok, I thought `_visual_` adds the same consistency as `_collision_` does.

PR #696, opened 2020-03-05
Fix analytic derivatives of Pinocchio; fix finite differencing on manifolds; updates to FDDP

wxmerkt:
This PR:

1. Fixes various dimensions in dynamic control problems, particular related to the size of the tangent space. It also exposes this variables explicitly (nq, nv, nu, nx, ndx). Further it reduces the number of heap allocations.
2. Fixes the analytic derivative of the `exotica_pinocchio_dynamics_solver` and eliminates allocations during runtime. Part of the partial derivative w.r.t. q was previously omitted.
3. Includes modifications to FDDP to make it (a) abstract and (b) to eliminate try-catch logic as identified in #692 
4. Fixes the default finite differencing in the dynamics solvers to correctly account for manifold operations, e.g., when using quaternions.

### DIFF
--- a/exotations/solvers/exotica_ddp_solver/src/feasibility_driven_ddp_solver.cpp
+++ b/exotations/solvers/exotica_ddp_solver/src/feasibility_driven_ddp_solver.cpp
@@ -120,7 +120,7 @@ void AbstractFeasibilityDrivenDDPSolver::Solve(Eigen::MatrixXd& solution)
     control_limits_ = dynamics_solver_->get_control_limits();
 
     AllocateData();
-    auto X_warm = prob_->get_X();
+    Eigen::MatrixXd X_warm = prob_->get_X();
     X_warm.col(0) = prob_->ApplyStartState();  // Apply start state
     auto U_warm = prob_->get_U();
     for (int t = 0; t < T_ - 1; ++t)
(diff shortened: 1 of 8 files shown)


exotations/solvers/exotica_ddp_solver/src/feasibility_driven_ddp_solver.cpp
```diff
@@ -211,7 +206,7 @@ void FeasibilityDrivenDDPSolver::Solve(Eigen::MatrixXd& solution)
             }
             catch (std::exception& e)
             {
-                // HIGHLIGHT("Error in TryStep: " << e.what())
+                WARNING_NAMED("NaN in ForwardPass", e.what())
```
VladimirIvan: Can the algorithm recover from NaNs here? Shouldn't this abort?

wxmerkt: It can recover in most cases, but I'd still like to see a warning and know.

VladimirIvan: Then it shouldn't use exceptions. Same reasoning as for FDDP and exceptions in a loop.

wxmerkt: Technically it is an exception and we do want to terminate the roll-out. It should not happen in a control loop with a well formulated problem but more during problem formulation development.

PR #697, opened 2020-04-01
Meshcat visualiser compatibility

VladimirIvan:
- Meshcat visualiser now follows mainstream Meshcat release
 - File serving is done vie a separate [server script](https://github.com/VladimirIvan/meshcat_ros_fileserver) that was split from [meshcat-python](https://github.com/rdeits/meshcat-python/pull/47). Install both scripts and start them using:
```
meshcat-server
meshcat-ros-fileserver -f /
```
 - If the file server is running on a different machine/port, you can specify this in the constructor of the `VisualizationMeshcat` class using a new argument `file_url`.

### DIFF
--- a/exotica_core/src/visualization_meshcat.cpp
+++ b/exotica_core/src/visualization_meshcat.cpp
@@ -29,6 +29,7 @@
 
 #ifdef MSGPACK_FOUND
 #include <chrono>
+#include <regex>
 
 #include <exotica_core/visualization_meshcat.h>
 #include <exotica_core/visualization_meshcat_types.h>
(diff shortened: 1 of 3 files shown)


exotica_core/src/visualization_meshcat.cpp
```diff
@@ -89,6 +90,16 @@ void VisualizationMeshcat::Initialize(bool use_mesh_materials)
     // https://github.com/rdeits/meshcat-python/blob/aa3865143120f5ace8e62aab71d825e33674d277/src/meshcat/visualizer.py#L60
     ConnectZMQ();
     web_url_ = RequestWebURL();
+    if (file_url_ == "")
+    {
+        std::regex url_regex("(.*):(?:\\d+)(?:\\/static\\/)");
+        std::smatch match;
+        if (std::regex_search(web_url_, match, url_regex) && match.size() > 1)
+        {
+            file_url_ = match.str(1) + ":9000/files/";
```
wxmerkt, line 99: This is hard-coded while the file server allows for it being configurable.

wxmerkt:
Awesome, thank you. Can you please add a page in the documentation with how to set it up and run it?

exotica_core/src/visualization_meshcat.cpp
VladimirIvan: This is just default behaviour. You can specify the full URL including the port to override it.

VladimirIvan:
I'll update the docker and add the docs in the near future.

PR #733, opened 2021-03-14
Add DistanceToLine2D task map

wxmerkt:
Allows defining a line from two frames and uses a third frame as a query point. Currently fixed to only accept one pairing. Unlike the default formula on Wikipedia we are not returning the absolute but the signed distance (to simplify the derivative).


wxmerkt:
I am planning to merge this in the afternoon tomorrow and then make a new binary release.

PR #735, opened 2021-04-03
[exotica_collision_scene_fcl_latest] Switch to FCL on Noetic, use fcl_catkin on older distros

wxmerkt:
ros-noetic-fcl has now been released into Noetic. This makes FCL 0.6.x
available directly via the ROS buildfarm and reduces the need for
fcl_catkin. Since we further cannot generate safe search paths
(exotica_core depends on moveit which depends on fcl.0.6 and this
library depends on fcl_catkin.0.6), replacing the use of fcl_catkin with
FCL on Noetic is the appropriate step to take. To keep backwards
compatibility with older distros (Melodic), we conditionally switch
between fcl_catkin and ros-noetic-fcl in this package.


wxmerkt:
This PR is blocked until upstream makes a new ros-noetic-fcl release as the current 0.6.1 version has bugs that cause segmentation faults on distance queries (including with our example script). Subsequently we can unrelease fcl_catkin for Noetic and switch to the upstream FCL version.

PR #755, opened 2023-10-19
Fix for visualising attached shapes

VladimirIvan:
Shapes created by defining custom links do not get published when calling publish_frames. This was because they were considered robot as robot geometry and got ignored. This avoids displaying the robot geometry twice, 1st as a part of the robot model and 2nd as the collision object published as a MarkerArray.
- Adds a separate logic to publish custom shapes attached to the robot but ignoring the shapes defined in the URDF.
- The links defined in the URDF are now correctly marked with `is_robot_link=true`. This is then used to distinguish links loaded from the URDF from custom links.

### DIFF
--- a/exotica_core/src/scene.cpp
+++ b/exotica_core/src/scene.cpp
@@ -809,6 +809,7 @@ void Scene::UpdateSceneFrames()
 
             std::string collision_element_name = links[i]->getName() + "_collision_" + std::to_string(j);
             std::shared_ptr<KinematicElement> element = kinematica_.AddElement(collision_element_name, trans, links[i]->getName(), links[i]->getShapes()[j]);
+            element->is_robot_link = true;
             model_link_to_collision_element_map_[links[i]->getName()].push_back(element);
 
             // Set up mappings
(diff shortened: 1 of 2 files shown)


VladimirIvan:
This wiil need rebasing after #756 is merged.

exotica_core/src/scene.cpp
```diff
@@ -783,6 +783,7 @@ void Scene::UpdateSceneFrames()
 
             std::string collision_element_name = links[i]->getName() + "_collision_" + std::to_string(j);
             std::shared_ptr<KinematicElement> element = kinematica_.AddElement(collision_element_name, trans, links[i]->getName(), links[i]->getShapes()[j]);
+            element->is_robot_link = true;
```
wxmerkt, line 786: Is this necessary? In the collision scene, we always check the `closest_robot_link.lock()` on the `weak_ptr` to check if it is attached to a robot link?

wxmerkt: Attached objects were always handled via the `closest_robot_link` instead of setting it to `is_robot_link` itself

VladimirIvan:
I checked the robot link checks. My edits were indeed not necessary.
I removed the changes and rebased the branch and this PR doesn't bring any useful changes. I'm closing this now

